PR #32, opened 2020-03-04
Simplify main(), magic() methods.

philippeitis:
I reduced the nesting of the inner code by using early returns / continues - this makes the code easier to read. Additionally, since magic_commands is a list, we can just check if command / args_given[0] in magic_commands, which makes the code a lot simpler. I also changed the `except Exception` line to `except KeyError`, which is the error that would be expected.

### DIFF
--- a/jc/cli.py
+++ b/jc/cli.py
@@ -185,66 +185,65 @@ def json_out(data, pretty=False):
 
 def magic():
     """Parse with magic syntax: jc -p ls -al"""
-    if len(sys.argv) > 1 and not sys.argv[1].startswith('--'):
-        parser_info = about_jc()['parsers']
-        # correctly parse escape characters and spaces with shlex
-        args_given = " ".join(map(shlex.quote, sys.argv[1:])).split()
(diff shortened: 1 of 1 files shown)


kellyjonbrazil:
Very nice!

I like the way you did the options and you taught me the .extend() method, which I wish I knew about earlier. :)

Much cleaner. I'll do some testing with it and pull it in if all is good!

kellyjonbrazil:
merged into `dev`. fixed up a couple minor bugs. still need to troubleshoot the case for no command (like `jc -a`)

```
$ jc -a
  File "/home/kbrazil/.local/bin/jc", line 11, in <module>
    load_entry_point('jc', 'console_scripts', 'jc')()
  File "/home/kbrazil/git/jc/jc/cli.py", line 254, in main
    magic()
  File "/home/kbrazil/git/jc/jc/cli.py", line 228, in magic
    elif args_given[0] in parser['magic_commands']:
IndexError: list index out of range
```

PR #33, opened 2020-03-04
Handle case where only options are passed.

philippeitis:
Arguments can appear like so: `-*`, `--*`, or `*`. Since we pop all args that match `-*` and return if `--*` exists, all that is left is keywords that match `*` - if these are not present, then the only thing that was in args_given was options, and so we should return.

### DIFF
--- a/jc/cli.py
+++ b/jc/cli.py
@@ -209,6 +209,10 @@ def magic():
             else:
                 break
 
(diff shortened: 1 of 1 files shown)


kellyjonbrazil:
looks like this works, thanks!

PR #34, opened 2020-03-04
Simplify process() in history.py, avoid list allocation in parse()

philippeitis:
Instead of doing two passes over the data in process(), it is possible to do one pass, which makes the code a lot simpler - using .isdigit() also helps avoid additional exception checks.

 In parse(), we don't use cleandata anywhere else, so we don't need to allocate a list, since filter() is iterable. This helps avoid a list allocation, and if all strings are empty, we will not enter the loop for filter - for instance:

```
a = ["", '', "", "hello"]
for word in filter(None, a):
    print(word)
>>> hello
```

You can iterate over the filter directly for almost every parser, which would help simplify and speed the code up since you would avoid a list allocation each time.

### DIFF
--- a/jc/parsers/history.py
+++ b/jc/parsers/history.py
@@ -79,21 +79,11 @@ def process(proc_data):
     # rebuild output for added semantic information
     processed = []
     for k, v in proc_data.items():
-        proc_line = {}
-        proc_line['line'] = k
-        proc_line['command'] = v
+        proc_line = {
+            'line': int(k) if k.isdigit() else None,
+            'command': v,
+        }
         processed.append(proc_line)
-
-    for entry in processed:
-        int_list = ['line']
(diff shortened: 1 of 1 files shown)


kellyjonbrazil:
cool - should probably also apply this to `foo.py` which is the template for new parsers. I can go back and fix the other parsers later, too.

kellyjonbrazil:
Tests pass. Merging... Thanks!

kellyjonbrazil:
didn't mean to merge this into master. wanted to merge into dev

kellyjonbrazil:
Sorry, i'm a little new to working with PRs. :) could you resubmit to `dev`. I should have done it but I missed it and now it doesn't seem easy to fix on my side after I reverted from `master`.

Thanks!

PR #37, opened 2020-03-05
Enable Continuous Integration with GitHub Actions.

philippeitis:
This automatically runs unit tests on various operating systems and Python versions when Python files are modified in pull requests or pushes to ensure that functionality remains correct without having to manually run tests each time, which should make the whole process easier and faster.

I've changed a few files - this is strictly to get CI tests to run - it also seems that the Windows tests do not pass due to failures with lsblk.py and stat.py

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -1,0 +1,31 @@
+name: Test code
+
+on:
+  push:
+    paths:
+    - "**/*.py"
+  pull_request:
+    paths:
+    - "**/*.py"
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8]
+
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


kellyjonbrazil:
Sweet - checking into this. Looks pretty straightforward.

I'll have to investigate why windows is failing. I've actually never tried python on windows. :)

philippeitis:
Looks like utf-8 errors - adding this forces the system to open the files in utf-8 mode without requiring further modifications, but is not ideal. 

```
_open = open
def open(*args, **kwargs):
    kwargs["encoding"] = "utf-8"
    return _open(*args, **kwargs)
```

kellyjonbrazil:
Ah yes, I do see some funky characters in those test files. Since these particular tests are more unixy in nature I wonder if it makes sense to skip them somehow in the github actions or do something like you suggested.

I wonder if just adding `encoding='utf-8'` to the open functions would fix it? If so, i'm not against going back and adding that to all of the tests myself.

philippeitis:
That'll work fine - I only modified the affected files, since updating all of the tests would be quite a large PR.

kellyjonbrazil:
Excellent - no worries I can go back and fix my mistakes. Really appreciate your contributions!

Issue #39, opened 2020-03-08
Footers not skipped in systemctl list-unit-files parser


kellyjonbrazil:
Footer lines not being skipped
```
...
  {
    "unit": "systemd-tmpfiles-clean.timer",
    "file": "static"
  },
  {
    "unit": "246",
    "file": "unit",
    "state": "files"
  }
]
```

kellyjonbrazil:
Looks like the issue is not with the parser, but with the Magic syntax.
```
jc -p systemctl list-unit-files
```
Looks like the Magic syntax is picking the wrong parser, probably plain `systemctl`.

kellyjonbrazil:
@philippeitis it looks like this issue was introduced https://github.com/kellyjonbrazil/jc/pull/32.

kellyjonbrazil:
fix merged into v1.8.1

PR #43, opened 2020-03-08
Fix #38, #39

### LINKED ISSUE #38: Footers not skipped in systemctl list-sockets parser
Footer lines are not being skipped.
```
…
 {
    "listen": "kobject-uevent",
    "unit": "1",
    "activates": "systemd-udevd-kernel.socket"
  },
  {
    "listen": "12",
    "unit": "sockets",
    "activates": "listed."
  },
  {
    "listen": "Pass",
    "unit": "--all",
    "activates": "to"
  }
]
```

philippeitis:
It seems that this magic() by matching the last valid parser in parser_info(), which seems problematic since it depends on the parsers being correctly sorted, which isn't validated. In this case, it seems like you may want to consider using a dictionary for your parser_info to do a simple lookup for the parser. -eg. 
```
if command in parser_info:
    found_parser = ...
elif args_given[0] in parser_info:
    found_parser = ...
```

### DIFF
--- a/jc/cli.py
+++ b/jc/cli.py
@@ -228,7 +228,7 @@ def magic():
                 pass
 
         # second pass for one word commands: e.g. 'ls'
-        elif args_given[0] in parser['magic_commands']:
+        if not found_parser and args_given[0] in parser['magic_commands']:
             try:
                 found_parser = parser['argument']
             # No command found - use standard syntax (for cases like 'jc -a')
--- a/jc/cli.py
+++ b/jc/cli.py
@@ -223,7 +223,6 @@ def magic():
         if command in parser['magic_commands']:
             try:
                 found_parser = parser['argument']
-                break
             # No command found - go to next loop (for cases like 'jc -a')
             except KeyError:
                 pass
@@ -232,7 +231,6 @@ def magic():
         elif args_given[0] in parser['magic_commands']:
             try:
                 found_parser = parser['argument']
-                break
             # No command found - use standard syntax (for cases like 'jc -a')
             except KeyError:
                 return


kellyjonbrazil:
Looks like this is working, thanks!

philippeitis:
And an alternative implementation which uses a dictionary:
```
def magic():
    """Parse with magic syntax: jc -p ls -al"""
    if len(sys.argv) <= 1 or sys.argv[1].startswith('--'):
        return

    magic_dict = {}
    parser_info = about_jc()['parsers']
    # Create a dictionary to of magic_commands to their respective parsers.
    for parser in parser_info:
        if "magic_commands" in parser:
            magic_dict.update({magic_command: parser for magic_command in parser["magic_commands"]})

    # correctly parse escape characters and spaces with shlex
    args_given = " ".join(map(shlex.quote, sys.argv[1:])).split()
    options = []

    # find the options
    p = 0
    for i, arg in list(enumerate(args_given)):
        # parser found - use standard syntax
        if arg.startswith('--'):
            return
        # option found - populate option list
        elif arg.startswith('-'):
            options.append(args_given.pop(i - p)[1:])
            p += 1
        # command found if iterator didn't already stop - stop iterating
        else:
            break

    if len(args_given) == 0:
        return

    # find the command and parser
    command = ' '.join(args_given[0:2])
    # Try to get a parser for command, otherwise get one for args_given[0]
    parser = magic_dict.get(command, magic_dict.get(args_given[0]))
    # Get found_parser if a parser exists.
    found_parser = parser.get("argument") if parser else None
    
    # construct a new command line using the standard syntax: COMMAND | jc --PARSER -OPTIONS
    run_command = ' '.join(args_given)
    if found_parser:
        cmd_options = '-' + ''.join(options) if options else ''
        whole_command = ' '.join([run_command, '|', 'jc', found_parser, cmd_options])
        os.system(whole_command)
    else:
        helptext(f'parser not found for "{run_command}"')
        sys.exit(1)
```

kellyjonbrazil:
I like the dictionary approach - it seems cleaner. I'll take a closer look at it and play around with it. Thanks!

kellyjonbrazil:
I took your dictionary approach and slightly modified it and it seems to work well. I streamlined `magic_dict` so it simply contains the command as the key and the argument as the value. A couple other minor changes to simplify the code and increase readability. Still need to test this on linux... only tested on my Mac, which doesn't allow testing for `systemctl`. Let me know if you have any concerns about this.
```
def magic():
    """Parse with magic syntax: jc -p ls -al"""
    if len(sys.argv) <= 1 or sys.argv[1].startswith('--'):
        return

    magic_dict = {}
    parser_info = about_jc()['parsers']

    # Create a dictionary of magic_commands to their respective parsers.
    for entry in parser_info:
        # Update the dict with all of the magic commands for this parser, if they exist.
        magic_dict.update({mc: entry['argument'] for mc in entry.get('magic_commands', [])})

    # correctly parse escape characters and spaces with shlex
    args_given = " ".join(map(shlex.quote, sys.argv[1:])).split()
    options = []

    # find the options
    popped = 0
    for i, arg in list(enumerate(args_given)):
        # parser found - use standard syntax
        if arg.startswith('--'):
            return

        # option found - populate option list
        elif arg.startswith('-'):
            options.append(args_given.pop(i - popped)[1:])
            popped += 1

        # command found if iterator didn't already stop - stop iterating
        else:
            break

    # all options popped and no command found - for case like 'jc -a'
    if len(args_given) == 0:
        return

    # find the command and parser
    one_word_command = args_given[0]
    two_word_command = ' '.join(args_given[0:2])

    # Try to get a parser for two_word_command, otherwise get one for one_word_command
    found_parser = magic_dict.get(two_word_command, magic_dict.get(one_word_command))

    # construct a new command line using the standard syntax: COMMAND | jc --PARSER -OPTIONS
    run_command = ' '.join(args_given)
    if found_parser:
        cmd_options = '-' + ''.join(options) if options else ''
        whole_command = ' '.join([run_command, '|', 'jc', found_parser, cmd_options])
        os.system(whole_command)
        exit()

    else:
        helptext(f'parser not found for "{run_command}"')
        sys.exit(1)
```

philippeitis:
I think it should be possible to add unit tests to actually run the commands on Linux if you're willing to look into installing the respective tools through a Github actions instance and then running automated tests against them and reading their input in to validate that they're correct.

kellyjonbrazil:
I agree - was thinking about this. Will have to find commands that will have consistent output, though. Not sure how easy that will be to do in practice.

kellyjonbrazil:
or maybe mock the commands? (make a systemctl list-unit-files command that outputs mock data) But not sure how easy it will be to put that into the path to override the real command.

philippeitis:
I might even move the core magic logic out into it's own function that generates a command / returns None, and then have a subprocess call in main() - this is more direct and allows easy testing regardless of system.

kellyjonbrazil:
Yes, that makes sense, so we are just unit testing the string output of magic() instead of integration testing the whole parsed output. Ok, I'll work on that. Thanks for the tip!

PR #44, opened 2020-03-08
Move core magic() logic into seperate function for testability, minor tweaks

philippeitis:
We only care about the command when testing magic() - by moving that out, we can easily test the command. I created a generate_magic_command() function, which returns a boolean and a command - the boolean signifies a valid command, and the command is a command or None if no command could be generated.

I also made some small tweaks (removed a list() call, fixed a possible bug with no arguments., moved magic_dict instantiation past the fast path checks to avoid making a dict if not needed.) This also includes two tests for the commands that didn't function correctly.

### DIFF
--- a/jc/cli.py
+++ b/jc/cli.py
@@ -183,18 +183,15 @@ def json_out(data, pretty=False):
         print(json.dumps(data))
 
 
-def magic():
-    """Parse with magic syntax: jc -p ls -al"""
+def generate_magic_command():
+    """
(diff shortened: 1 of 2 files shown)


kellyjonbrazil:
very nice. I tried to commit back to this PR, but it created a new PR/44 branch where I added some tests and minor tweaks.

philippeitis:
I think it's also fine to just pull against the patch and I can merge it manually - but I think it should be possible for you to commit directly since I do allow edits from maintainers.

PR #65, opened 2020-06-06
Load custom parsers from <user_data_dir>/jc/jcparsers

duelafn:
Hello, I'd like to propose a feature for custom (local) parsers. Benefits:

* Lower the barrier for entry to write new parsers
* Allow override of builtin parsers for strange systems or command-line arguments
* Allow use of jc with uncommon or specialized apps (for which upstream submission would not be appropriate)

This PR includes such a feature and will include parsers present in `.py`
files placed into the `jc/jcparsers` folder of a user's app-data directory.
Determining an appropriate app-data directory for the current OS did
require a new dependency on the "appdirs" package.

I have tested the changes on Linux. I don't have a Windows or Mac available
to test or verify the path locations I documented, but these paths seem to
be correct judging from the appdirs documentation. Aside from possibly
mis-documenting the correct extension folders, I do not believe there are
any OS-specific dangers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,13 @@ Tested on:
 - NixOS
 - FreeBSD12
 
+
+## Custom Parsers
+Custom local plugins may be placed in a "jc/jcparsers" folder in your local "App data directory" (For example, "/home/*myname*/.local/jc/jcparsers" on Linux, "/Users/*myname*/Library/Application Support/jc/jcparsers" on Mac OS X, or "C:\Users\*myname*\AppData\Local\jc\jc\jcparsers" on Windows.
+
+Use the [`jc/parsers/foo.py`](https://github.com/kellyjonbrazil/jc/blob/master/jc/parsers/foo.py) parser as a template and simply place a `.py` file in the `jcparsers` subfolder. Local plugin file names must be valid python module names, therefore must consist entirely of alphanumerics and start with a letter. Local plugins may override default plugins.
+
+
 ## Contributions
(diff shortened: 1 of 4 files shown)


kellyjonbrazil:
Very nice! Thank you for the contribution. I like the idea and had thought a bit about it but hadn’t even thought about how to implement it. Let me digest this over the next couple days. Looks like a great feature and a slick implementation!

kellyjonbrazil:
Did some quick testing on OSX. I like how it works - nicely overrides existing parsers and adds the information to jc help and `jc -a` just fine.

My initial concern was with the appdirs requirement as jc is currently being packaged on more platforms. It seems that appdirs is available on the distros I checked (nixos, freebsd ports, fedora, debian, etc.) so it looks like this dependency shouldn't be a problem. I didn't see it in Homebrew, but the way those apps are packaged, I don't think it matters, we'd just add appdirs to the dependency list and point to the file on pypi.

So I think we are ok with the requirement. Alternatively we could vendorize it since it's so small, and it specifically states vendorizing appdirs is a use case in their code. Do you have a strong opinion either way?

I actually wouldn't mind vendorizing it so we make the downstream packagers' lives easier and we can switch to the external library later if we run into compatibility issues with future operating systems or versions.

duelafn:
I have no strong opinion about dependency vs vendorized appdirs.

kellyjonbrazil:
Cool, I'll go ahead and merge this into Dev. I'll probably vendorize appdirs before merging into master and publishing.  Thanks again for the PR - great stuff!

kellyjonbrazil:
fyi, this has been added to v1.11.3. Thanks!

PR #72, opened 2020-07-09
Support older pygments

duelafn:
Tested/works on pygments 2.3.1, this backward support may help your Debian request for packaging since it makes you independent of a pygments update.

Feel free to reject this request if you don't want the legacy baggage.

### DIFF
--- a/jc/cli.py
+++ b/jc/cli.py
@@ -17,7 +17,6 @@ from pygments.style import Style
 from pygments.token import (Name, Number, String, Keyword)
 from pygments.lexers import JsonLexer
 from pygments.formatters import Terminal256Formatter
-from packaging import version
 import jc.appdirs as appdirs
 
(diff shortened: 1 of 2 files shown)


kellyjonbrazil:
It's funny - I was just starting to look into this since the Debian pygments version is so old and I haven't seen any movement on upgrading it. My concern is with the packaging library as I'm trying to keep from adding more dependencies.

Since we're just comparing the versions and pygments seems to use canonical versioning at least since v2.3.0, I wonder if a simple tuple compare will work instead of using the full-blown version function?

kellyjonbrazil:
Nice, thanks! I'll merge this. I'll probably continue working on 1.12.0 for the next couple weeks, maybe longer. I want to add more tests for cli.py and add a couple more parsers.

PR #97, opened 2021-01-06
iw_scan: Improve detection of lines starting with spaces

pschmitt:
Ref https://github.com/kellyjonbrazil/jc/issues/94#issuecomment-755162509

This improves the detections of lines that start with space chars in the iw_scan parser.

Without this one only gets the `mac_address` and `interface` fields on Arch Linux and Termux.

### DIFF
--- a/jc/parsers/iw_scan.py
+++ b/jc/parsers/iw_scan.py
@@ -109,6 +109,7 @@ Examples:
       ...
     ]
 """
+import re
 import jc.utils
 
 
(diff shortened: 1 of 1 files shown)


kellyjonbrazil:
Do you mind if I credit you in the parser `info` class? This will show up in `jc -a`.

pschmitt:
I hereby grant you my explicit permission to do that ;)

PR #106, opened 2021-03-25
Add windows systeminfo command parser

jon-rd:
- Adds parsing of Windows "systeminfo" output

Related to [Windows command support discussion](https://github.com/kellyjonbrazil/jc/discussions/105)


kellyjonbrazil:
Nice work! I'll take a closer look, but a couple things I forgot to mention in the previous discussion:

- As much as possible I try to normalize keys so they are always lowercase (like you have done), but also replace spaces with `_` and maybe just remove parenthesis. This way no special tricks have to be used when filtering with tools like `jq`
- I would make the parser version 0.5 (beta) instead of a 1.0 if there are no additional processing of values (e.g. convert strings to ints/floats). This way the user can expect that the dict/json may change in a future version if those values get additional processing. Some parsers don't require additional processing, but this type of command output is a good candidate for that type of thing.

Also, just curious as I've never actually tested this on Windows - does the magic syntax actually work? `jc -p systeminfo`

Thanks again for the contribution!

jon-rd:
Thank you for the quick feedback! The changing of the key names to remove special chars was such a big change that I just chose to overwrite the original commit with the update. However, I added a second commit to address your followon feedback regarding the version (as well as added the parser to cli.py and added a line to the docgen.sh file).

And thank you for clarifying the magic syntax. I had not tested it before, but I ran it just now on a Windows box with the following output:

```
C:\Users\Administrator\git\jc>jc -m -p systeminfo
{
	"host_name": "WIN-1A11AAAAAAA",
	"os_name": "Microsoft Windows Server 2019 Standard Evaluation",
	"os_version": "10.0.17763 N/A Build 17763",
	<snip>
}
```

jon-rd:
One more thing - I didn't run docgen.sh. So one more commit to add in the generated systeminfo.md file.

kellyjonbrazil:
Hi there - I was looking through the `systeminfo` output in more detail and I have to say this is a pretty ambitious parser, but it is doable. Since the output structure is fairly stable I think the schema should better reflect the nesting of the data displayed. Something like this:
```
{
   "host_name": "DESKTOP-WIN01",
   "os_name": "Microsoft Windows 10 Enterprise",
   "os_version": "10.0.19042 N/A Build 19042",
   "...": "...",
   "processors": [
      "Intel64 Family 6 Model 158 Stepping 13 GenuineIntel ~2400 Mhz"
   ],
   "bios_version": "VMware, Inc. VMW71.00V.11111111.B64.2008100111, 8/10/2020",
   "...": "...",
   "total_physical_memory_mb": 2047,
   "available_physical_memory_mb": 1417,
   "...": "...",
   "hotfixes": [
      "KB4578968",
      "KB4562830",
      "KB4570334",
      "KB4580325",
      "KB4586864",
      "KB4594440"
   ],
   "network_cards": [
      {
         "name": "Intel(R) 82574L Gigabit Network Connection",
         "connection_name": "Ethernet0",
         "dhcp_enabled": true,
         "dhcp_server": "192.168.133.250",
         "ip_addresses": [
            "192.168.133.3",
            "fe80::192:eb64:1fcf:86eb"
         ]
      }
   ],
   "hyper_v_requirements": {
      "vm_monitor_mode_extensions": true,
      "virtualization_enabled_in_firmware": true,
      "second_level_address_translation": true,
      "data_execution_prevention_available": true
   }
}
```
(`hyper_v_requirements` would be `null` if those don't show up.)

This output reminds me of `sysctl`, `dmidecode`, and some others where I had to do some ugly hacks to get the fields correct. If you feel up for it I'd love to see something more like the above. If not, I can probably tackle this in the next couple weeks.

kellyjonbrazil:
Hey, just checking in to see how things are going. I'm starting to wrap things up for the next 1.15.0 `jc` release and I'd love to include the `systeminfo` parser in the release.

If you'd like to pass on the proposed schema I can take a try at it. It would be great to get some more example `systeminfo` outputs for testing. I found a couple on the web, but it would be great to have more samples.

Let me know what works for you.

Thanks!

jon-rd:
Hey! I've rebased on latest in dev. Current state:

- Hotfixes: Added parsing for it. One note that I made in the code is that it seems like if there's a high number of hotfixes, that the command just cuts off the list after a certain number. If you look at the Windows 7 sample, you'll see  that it says ~300 hotfixes installed, but the list stops after 245 of them.
- Processors: Added parsing for it
- Network information: I'd like to gather more samples before I try to parse the network information though as its schema has alot of variety between the two samples that I've gathered so far.
- HyperV Requirements - I haven't setup a valid hyper-v workstations to see the output from it. Willing to do so as soon as I gather it.

Thoughts welcome.

kellyjonbrazil:
Looking good! Yeah, the difficult part sometimes is getting enough samples to feel confident in the parsing. I think the network info/ip address info will be very valuable in scripts. Looking forward to it.

kellyjonbrazil:
Hi there - I'm ready to publish v1.15.0 now. Do you need a couple more days to finish up or should I go ahead and we can add the `systeminfo` parser in 1.15.1? I'm open either way. Thanks!

jon-rd:
Go ahead and release. I'm gathering "systeminfo" from a couple more machines and adding the discussed parsing, but I'm not going to be able to finish the work until Friday. Thanks!

kellyjonbrazil:
Hey - all set for me to take a look? From what I've seen so far it looks great!

jon-rd:
Yes! Let me know what I've missed, and I should be able to get any needed changes back to you quickly.

Side note: I'm not sure I fully understand how you handle local testing of outputs that contain timestamps. Tests that contain expected timestamps fail when I run them locally but pass when run via github actions. I'm assuming this is because my local timezone is different than the github actions timezone. How do you run tests locally without changing your system time zone?

kellyjonbrazil:
Thanks - regarding timestamps I have github actions systems set to `America/Los_Angeles` for linux/macOS and `Pacific Standard Time` for Windows to match my local system. The timestamps are pretty tricky, so the `naive` ones will pretty much never match unless you happen to be set to the same timezone, while the `utc` timestamps should always be the same.

I think your PR is in good enough shape that I can merge it and make some minor tweaks before publishing in v1.15.2.

kellyjonbrazil:
Beautiful parser! I have it merged into dev and made some minor updates. Thanks for the contribution.

kellyjonbrazil:
@jon-rd  Just a heads-up, the latest version of `jc` (v1.15.4) now supports colors in Windows.

kellyjonbrazil:
@jon-rd Also another heads-up, I was able to package `jc` v1.15.4 as an MSI package:
https://github.com/kellyjonbrazil/jc/releases/tag/v1.15.4

PR #107, opened 2021-03-25
Add windows dir command parser

rasheed-rd:
- Adds parsing of Windows "dir" output

Related to [Windows command support discussion](https://github.com/kellyjonbrazil/jc/discussions/105)


jc/parsers/dir.py
```diff
@@ -0,0 +1,332 @@
...
+                continue
+
+            output_line = {}
+            parsed_line = line.split()
+            output_line["date"] = parsed_line[0]
+            output_line["time"] = " ".join(parsed_line[1:3])
+            output_line.setdefault("dir", None)
+            output_line.setdefault("size", None)
+            if parsed_line[3] == "<DIR>":
+                output_line["dir"] = parsed_line[3]
```
kellyjonbrazil, line 322: I wonder if it makes sense to turn this field into a boolean so if "<DIR>" is detected, then output_line['dir'] = True ? Otherwise, False. This way it's easy to programmatically tell if this is a directory or a file from the one field.

kellyjonbrazil:
Very, nice - thanks for contributing! I'll take a closer look soon.

Just from my experience with the `ls` parser, did you happen to test scenarios where you `dir` a single file vs multiple files and a directory vs multiple directories? I'm not as familiar with `dir` output, but `ls` would give slightly different output depending on how the command was constructed and I had to account for that, otherwise the field mappings could change. Not sure if that affects `dir` in a similar way.

jc/parsers/dir.py
rasheed-rd: yeah that makes more sense. I'll update it.

rasheed-rd:
I did. I tested single and multiple files and directories. I also tested a mix of both. I'll create a test case b/c it does change the output a bit.

tests/fixtures/windows-10/dir-C.out
```diff
@@ -0,0 +1,21 @@
+ Volume in drive C has no label.
+ Volume Serial Number is 246B-3B5A
+
+ Directory of C:\Program Files\Internet Explorer
```
kellyjonbrazil, line 4: Might be interesting to pull `C:\Program Files\Internet Explorer` into a `parent_directory` field in each `output_line`.

tests/fixtures/windows-10/dir-Q.out
```diff
@@ -0,0 +1,21 @@
...
+03/24/2021  03:15 PM                 0 BUILTIN\Administrators file name.txt
+12/07/2019  02:09 AM            54,784 NT SERVICE\TrustedInstahmmapi.dll
+12/07/2019  02:09 AM           515,072 NT SERVICE\TrustedInstaiediagcmd.exe
+12/07/2019  02:09 AM           504,832 NT SERVICE\TrustedInstaieinstal.exe
+12/07/2019  02:09 AM           224,768 NT SERVICE\TrustedInstaielowutil.exe
+12/07/2019  02:09 AM           421,888 NT SERVICE\TrustedInstaIEShims.dll
+12/06/2019  02:47 PM           819,136 NT SERVICE\TrustedInstaiexplore.exe
+12/07/2019  02:14 AM    <DIR>          NT SERVICE\TrustedInstaimages
+12/07/2019  02:14 AM    <DIR>          NT AUTHORITY\SYSTEM    SIGNUP
+12/07/2019  02:09 AM            48,536 NT SERVICE\TrustedInstasqmapi.dll
```
kellyjonbrazil, line 19: The `dir /Q` output looks quite hard, if not impossible, to parse correctly. Maybe we should just specify that `/Q` is not supported or just don't include it as one of the supported options. People can still try to use it at their own risk, but it won't be 'supported'

kellyjonbrazil:
Hi there, I'm getting pretty close to wrapping v1.15.0 and I'd like to include this parser, if possible. It looks pretty good and I can clean up anything else that is needed, but could you provide a few more samples that I can add to the tests?

Thanks!

rasheed-rd:
Hi, sorry, I've been swamped for about a week didn't get a chance to work on this until yesterday. I was able to include the parent dir which changed the structure a bit so I'm having to update all the test samples. I should have it all pushed today.

rasheed-rd:
The following updates were made:
- Made the dir field a bool
- Added parent dir structure
- Removed support for /Q
- Added support for /S
- Added several new test cases

Looking forward to any feedback you may have.

kellyjonbrazil:
Great stuff! One thing, though, is I think instead of having a parent directory key with an array of files inside it should just be a flat array of files with a parent directory field inside each one.

I know it sounds strange, but there are a couple reasons for this:
- (less importantly) this will mimmic the `ls` parser
- As much as possible, I like to have key names be static. This makes it easier to search through the records with tools like `jq`
- As much as possible, I like to have key names be lowercase, no spaces, and with no special characters.

For example, if a user needs the data for a certain file, if it's just a flat list of objects they can easily iterate through it and grab the ones they need using a `select` function or whatever. If they key name is variable, it not only makes it more complex to search, but the path to the file information becomes longer and less predictable:
```
.["C:\\Program Files\\Internet Explorer"][0].filename
```
vs.
```
.[0].filename
```
If you don't mind, I can take what you have and make the changes. I understand schema debates can be quite subjective, so if you have strong feelings on the matter, feel free to let me know. :)

rasheed-rd:
That makes sense. No objections!

kellyjonbrazil:
@rasheed-rd Just a heads-up, the latest version of `jc` (v1.15.4) now supports colors in Windows.

rasheed-rd:
Awesome!

kellyjonbrazil:
@rasheed-rd Also another heads-up, I was able to package jc v1.15.4 as an MSI package:
https://github.com/kellyjonbrazil/jc/releases/tag/v1.15.4

Issue #158, opened 2021-08-24
Parser for generic ASCII tables.


CtrlC-Root:
I have a few tools I use that output their data as ASCII tables in various formats. An example would be the `virsh` commnad line interface:

```
$ virsh pool-list --all                                                                                                                                            
 Name          State    Autostart
-----------------------------------
 default       active   no
 disk-images   active   yes
 vmdisk        active   yes
```

The issue is that `virsh` alone has hundreds of commands which output different columns. Some of those commands even allow selectively hiding or including specific columns in the output.

I'd like to parse the output of these but I don't know that it makes sense to write a parser for each one or even for each of these programs. What if we added a generic ASCII table parser instead that could handle some common output formats? As an example here's a short Python 3 script I wrote as a first pass at this: https://github.com/CtrlC-Root/dotfiles/blob/master/bin/tj. Would something like this make sense to add to this project?

kellyjonbrazil:
Interesting - I do have a `universal` parser that works on columnar tables output like `ps` and `netstat`. It's an internal parser used by the higher-level parsers:

https://github.com/kellyjonbrazil/jc/blob/master/jc/parsers/universal.py

A higher-level ASCII Table parser could probably be built on top of one or both of the Universal parsers.

kellyjonbrazil:
Hi @CtrlC-Root any thoughts on bringing this back to life? I think this would be a cool parser! If you don't have plans to pick this up, maybe I'll take a stab at it.

Thanks!

kellyjonbrazil:
I have a couple of ASCII table parsers working in `dev` now.

The first is the standard, single line table parser and it can handle all sorts of ASCII and Unicode tables:
```
$ echo '                                                                                                                                           
>  Name          State    Autostart
> -----------------------------------
>  default       active   no
>  disk-images   active   yes
>  vmdisk        active   yes' | jc --asciitable -p
[
  {
    "name": "default",
    "state": "active",
    "autostart": "no"
  },
  {
    "name": "disk-images",
    "state": "active",
    "autostart": "yes"
  },
  {
    "name": "vmdisk",
    "state": "active",
    "autostart": "yes"
  }
]
```
Markdown tables:
```
$ echo '
>     | type   |   total |   used |    free |   shared |   buff cache |   available |
>     |--------|---------|--------|---------|----------|--------------|-------------|
>     | Mem    | 3861332 | 222820 | 3364176 |    11832 |       274336 |     3389588 |
[...]

PR #167, opened 2021-09-09
Implement generic ASCII table parser

CtrlC-Root:
This is based on the discussion in #158. In general this seems to work but there's a few rough edges I'd like feedback on:

* Some tables use separator characters to denote columns. This implementation doesn't handle that.
* Some tables use column headers that contain multiple words while others use single words. This implementation detects headers using at least two spaces which handles the former correctly (i.e. `virsh domifaddr foo`) but fails for the latter (i.e. `df` although there's already a dedicated parser for that command so it's only an example here).

I tried to find a way to use the `jc.parsers.universal.sparse_table_parse` as the underlying implementation but unless I've misread it's implementation I think it cannot correctly handle all the cases listed above. If we want to be able to handle ASCII tables that use separator characters between columns or one or more words per column header then I think we have to look at all of the rows to correctly pick out columns (i.e. by detecting a column of spaces or identical separator characters). Still this implementation gets most of the way there without a lot of code.


jc/parsers/ascii.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # compatible options: linux, darwin, cygwin, win32, aix, freebsd
+    compatible = ['linux', 'darwin', 'cygwin', 'win32', 'aix', 'freebsd']
+    magic_commands = ['ascii']
+
+
+__version__ = info.version
+
+
+def _process(proc_data):
```
CtrlC-Root, line 55: I can't think of anything I could do in this function for generic tables. Should I remove it and ignore the `raw` parameter to `parse`?

jc/parsers/ascii.py
```diff
@@ -0,0 +1,178 @@
...
+    
+        line:    (string) header row line data
+        
+    Returns:
+    
+        Line with cleaned header values.
+    """
+
+    # XXX: what about tables that only use a single space between columns?
+    matches = list(re.finditer(r"\w+( \w+)*", line))
```
CtrlC-Root, line 108: This regular expression is used to detect header values.

jc/parsers/ascii.py
```diff
@@ -0,0 +1,178 @@
...
+    
+        True if the line contains headers or data
+    """
+
+    # ignore empty or blank lines
+    if not line:
+        return False
+
+    # ignore separator lines (i.e. "----")
+    if len(set(line)) <= 1:
```
CtrlC-Root, line 141: Too clever or too simple?

jc/parsers/ascii.py
kellyjonbrazil: Yeah, that's fine. There are a couple other parsers where I basically call `_process` but all it does is return `proc_data` unchanged. I just do that in case I find a use for it in the future.

jc/parsers/ascii.py
kellyjonbrazil: Ha - I really like that. Very clever and elegant!

CtrlC-Root: Well good :)

kellyjonbrazil: Actually, maybe just one little tweak:
```
if len(set(line.replace(' ', ''))) <= 1:
```
This way it doesn't count spaces as part of the separator line. This is for cases like the output of `jtbl`:
```
$ jc df | jtbl -n
filesystem       512_blocks       used    available    iused                ifree  mounted_on           capacity_percent    iused_percent
-------------  ------------  ---------  -----------  -------  -------------------  -----------------  ------------------  ---------------
/dev/disk1s1      976490576  664388168    300216728  2679309  9223372036852096498  /                                  69                0
devfs                   660        660            0     1142                    0  /dev                              100              100
/dev/disk1s4      976490576   10485984    300216728        5  9223372036854775802  /private/var/vm                     4                0
```

CtrlC-Root: What if there's a table like this:

```
Column  Column
--------  ---------
-                -
X               X
```

Although I suppose you could argue that it's ok to exclude those rows.

It feels like there's no way to solve this problem for the general case when you're looking at one row at a time. To handle that we probably need to:

1. Looks at all the columns and picks the separators (based on matching values in each row) then determine the valid text range for each.
2. Looks at all the rows and picks out the separators (based on matching values in each column possibly ignoring the ranges used to separate columns as detected above) and keep the header and data rows.

The result of that could probably be fed into `sparse_table_parse` actually although we'd be reading the data at least 3 times then.

kellyjonbrazil:
I wonder if we can handle column separator characters with a regex that looks for the `|` character with a space before/after it or if the line begins/ends with it and a space:
```
(^[|]\s)|(\s[|]$)         # replace these with two spaces
```
```
(\s[|]\s)                    # replace these with 3 spaces
```
Not sure if there are other column separator characters we should be concerned with, but they can easily be added to the regex.

jc/parsers/ascii.py
kellyjonbrazil: I'll take a closer look at the regex itself, but any reason not to use `re.findall()` instead of `re.finditer()` since it is being cast to a list anyway?

CtrlC-Root: IIRC it's because `findall()` returns tuples without the start and end positions while `finditer()` returns match objects that do include those. Compare the output:

```
>>> re.findall(r'\w+( \w+)*', ' Hello World    This is cool    Test')
[' World', ' cool', '']
>>> list(re.finditer(r'\w+( \w+)*', ' Hello World    This is cool    Test'))
[<re.Match object; span=(1, 12), match='Hello World'>, <re.Match object; span=(16, 28), match='This is cool'>, <re.Match object; span=(32, 36), match='Test'>]
>>>
```

jc/parsers/ascii.py
kellyjonbrazil: Yeah, to do it right is going to be a bit of analysis. Part of me is cool with an MVP and let's see if any issues are reported. But it would be slick to do it right with a non-lossy parser. A lot more code, though.

jc/parsers/ascii.py
CtrlC-Root: I could probably take out the `list()` call though and iterate using the iterator returned directly. Think that was left over from debugging.

kellyjonbrazil: Interesting  - I'll have to wrap my head around why the output is different with the same regex pattern, but I see you are using the start/end attributes that come with `re.finditer` so that makes sense.

jc/parsers/ascii.py
```diff
@@ -0,0 +1,178 @@
...
+    """Provides parser metadata (version, author, etc.)"""
+    version = '1.0'
+    description = 'ASCII table parser'
+    author = 'Alexandru Barbur'
+    author_email = 'alex@ctrlc.name'
+    # details = 'enter any other details here'
+
+    # compatible options: linux, darwin, cygwin, win32, aix, freebsd
+    compatible = ['linux', 'darwin', 'cygwin', 'win32', 'aix', 'freebsd']
+    magic_commands = ['ascii']
```
kellyjonbrazil, line 49: You can go ahead and remove/comment out the `magic_commands` attribute since it is only used for magic syntax and there is no command for `jc` to run and use this parser.

kellyjonbrazil:
Is it cool if we move the merge from `master` to `dev`?

jc/parsers/ascii.py
kellyjonbrazil: Another thought to keep things simple - only check for the separator line right after the header if one is detected. Still wouldn't be perfect, but might cover 80%+ of the use cases.

CtrlC-Root:
> Is it cool if we move the merge from `master` to `dev`?

My bad, I meant to do that but I must have missed the option when I was creating it. I'm going to close this PR, make the changes you recommended, see if I can include a few more tests and maybe a better implementation, then open a new PR against `dev`.

kellyjonbrazil:
No worries - right on!

PR #185, opened 2021-12-19
initial commit zipinfo.py to new_branch

listuser:
Flattened schema
Tested on RHEL8 using Ansible 2.9 as well as cli

-Matt


kellyjonbrazil:
Very nice - thanks for the contribution!

Could you add some sample `zipinfo` output files under `tests/fixtures/...` that can be used to validate the parser? It looks like `zipinfo` is included on macOS, too - we might check it for that output, as well.

listuser:
zipinfo.out and zipinfo.json uploaded to tests/fixtures/rhel-8

You weren't specific about the directory and I didn't see a rhel-8, so I created it.

kellyjonbrazil:
Perfect, thanks! I'll take a look.

kellyjonbrazil:
Looks like it's working well. Though I typically like to keep the schema as flat as possible, in this case I think it leads to a lot of redundant data, so I think we should move the individual file records under a `files` array, so it will look like:
```
{
    "archive": "log4j-core-2.16.0.jar",        # string
    "size": "1789565",                                 # int
    "bytes_compressed": "1515455",        # int
    "bytes_uncompressed": "3974141",    # int
    "number_entries": "1218",                    # int (not sure if one of these is redundant)
    "number_files": "1218",                         # int (not sure if one of these is redundant)
    "size_unit": "bytes",                              # this string needs the comma stripped
    "percent_compressed", 61.9               # float,
    "files": [
        {
            "flags": "-rw-r--r--",                     # string
            "zipversion": "2.0",                       # string
            "zipunder": "unx",                        # string
            "size": "1789565",                       # int (this was getting overriden by the zipfile size)
            "type": "bl",                                  # string
            "method": "defN",                       # string
            "date": "21-Dec-12",                   # string
            "time": "23:35",                           # string
            "filename": "Log4j-events.xsd"   # string
        }
    ]
}
```
The int/float conversions can be done in the `process` function.

listuser:
I thought you might want a nested version which I already had prepared. Will commit soon.

listuser:
I will work on the the int/float conversions in the process function as time permits. Thanks for the feedback.
Of note, the nested version I just uploaded is slightly different from what you spec'd above.

kellyjonbrazil:
Yep, looks good, but I would even get rid of that `archive` key. I was checking around and I couldn't see where `zipinfo` would report on multiple archives. I couldn't get it to on the command line, so I think just putting all of the archive info as top level, and then put all the files under a `contents` array is fine. If multiple archives is possible, then we can nest the outermost object in another array, but like I said, I'm not sure `zipinfo` really supports that.

listuser:
That makes sense.
Will do along with the int/float conversions in the process function, as time permits.

listuser:
Regarding

    "number_entries": "1218",                    # int (not sure if one of these is redundant)
    "number_files": "1218",                         # int (not sure if one of these is redundant)

I haven't come across an archive where the numbers are different yet. Don't know if its redundant.
Will review your suggestions again.

kellyjonbrazil:
> I haven't come across an archive where the numbers are different yet. Don't know if its redundant. Will review your suggestions again.

All good - probably not a problem to leave them both now, especially if it's just being printed in the archive meta area once.

Nice work!

listuser:
schema updated, conversions completed

kellyjonbrazil:
Looks great!

I see you wrapped the object in an array - I think this makes sense in case we find a way for `zipinfo` to report on multiple archives. But if we are sure it doesn't, then we can probably remove it. I really don't know the answer, so this is probably the safest bet so we don't need to change the schema in a future version.

I'll go ahead and merge this into `dev`, add some tests and then release in jc v1.17.5.

Thanks for contributing!

listuser:
> I see you wrapped the object in an array 

I can revise the schema if you prefer to have the array removed, time permitting.

Your feedback has been helpful. I am fairly new to python and expected some back and forth regarding the parser and schema.

Let me know. 

BTW, I stumbled on your project through the Ansible docs as I was looking for a way to parse semi-structured text.

https://docs.ansible.com/ansible/latest/network/user_guide/cli_parsing.html#parsing-with-jc

kellyjonbrazil:
All good - I'll do some more testing on different systems (BSD, MacOS, Linux) and see if I can coax multiple archive support out of `zipinfo`. If not, i may just unwrap the schema to make it easier for the end-user.

I liked the tuple unpacking - nice!

listuser:
The zipinfo version on my linux system supports wildcard expressions similar to Unix egrep.

Wildcard needed to be escaped on my linux system as follows:

zipinfo \*.jar

Doing this

zipinfo \*.jar | jc --zipinfo

causes the zipinfo template to error. The output does not match the parsing.

kellyjonbrazil:
Yeah, I  tried that too, but using glob expansion seems to have `ziipinfo` interpret the first filename as the archive and then the rest of the filenames are for details about files within the archive. The error you hit is probably because a file with that name (the second file that comes up in the glob expansion) does not exist inside the archive, so the `zipinfo` output is missing information.

I wasn't able to get `zipinfo` to report on multiple archives, though the man page insinuates that it might be possible. I don't think it was ever implemented, but I'll keep checking.

listuser:
I revised my comment as I didn't realize I needed to backtick the commands. See above.

listuser:
I think checking the last line of output from a multi archive zipinfo (as above) will be sufficient to determine the parsing steps.

Last line output example
73 archives were successfully processed.

kellyjonbrazil:
Ah yes - that was the trick! Once I encapsulated the wildcard in quotes I got multiple archives:
```
$ zipinfo '*.zip'
Archive:  jc1.zip
Zip file size: 4116 bytes, number of entries: 1
-rw-r--r--  2.1 unx    11837 bX defN 21-Dec-08 20:50 jc.1
1 file, 11837 bytes uncompressed, 3966 bytes compressed:  66.5%

Archive:  testzip.zip
Zip file size: 8106 bytes, number of entries: 2
-rw-r--r--  3.0 unx      197 tx defN 21-Aug-03 15:12 round-table.gv
-rw-r--r--  3.0 unx     8342 bx defN 21-Aug-03 15:12 round-table.gv.pdf
2 files, 8539 bytes uncompressed, 7651 bytes compressed:  10.4%

Archive:  micro.zip
Zip file size: 6144 bytes, number of entries: 8
-rw-r--r--  2.1 unx    10688 bX defN 19-Sep-30 16:47 microsimservermac.py
drwxrwxr-x  2.1 unx        0 bx stor 21-Dec-20 14:33 __MACOSX/
-rw-r--r--  2.1 unx      176 bX defN 19-Sep-30 16:47 __MACOSX/._microsimservermac.py
-rw-r--r--  2.1 unx      528 bX defN 19-Aug-27 07:46 Dockerfile
-rw-r--r--  2.1 unx    10538 bX defN 19-Oct-01 13:22 microsimserver.py
-rw-r--r--  2.1 unx      380 bX defN 19-Oct-01 13:22 changelog.txt
-rwxr-xr-x  2.1 unx      263 bX defN 19-Oct-01 12:09 dockerhub.sh
-rw-r--r--  2.1 unx      266 bX defN 19-Oct-01 12:09 __MACOSX/._dockerhub.sh
8 files, 22839 bytes uncompressed, 4908 bytes compressed:  78.5%

3 archives were successfully processed.
```
I can fixup the parser to handle this - unless you want to work on it. Up to you. If so, it would be best to pull again from `dev` since I already merged it. Sorry I jumped the gun! :)

listuser:
If you don't mind, please do fixup the parser. As I am new to python, I would like to see how you go about resolving the problem. Thanks.

kellyjonbrazil:
I updated the parser to support multiple archives if you want to check it out.

listuser:
looks good. Thanks for fixing multi.

Noticed some unexpected screen output when using the parser.

without parser
```
apache-log4j-2.17.0-bin]$ zipinfo '*.jar' &>out
apache-log4j-2.17.0-bin]$
```

with parser
```
apache-log4j-2.17.0-bin]$ zipinfo '*.jar' | jc --zipinfo -p &>out

73 archives were successfully processed.
apache-log4j-2.17.0-bin]$
```

I didn't expect to see anything about the successfully processed.

kellyjonbrazil:
Yeah it seems `zipinfo` is really aggressive with that status message and writes it to both STDOUT and STDERR when there are multiple archives. Nothing really to worry about, but the message can be hidden by redirecting STDERR to `/dev/null` or something.

I noticed that when I was running the output through `cat` to see how `zipinfo` changes its output when being piped:
```
$ zipinfo '*.zip' | cat
zipinfo:  cannot find or open *.zip, *.zip.zip or *.zip.ZIP.

No zipfiles found.
zipinfo:  cannot find or open *.zip, *.zip.zip or *.zip.ZIP.

No zipfiles found.
kbrazil-mac:jc kbrazil$ cd ~
kbrazil-mac:~ kbrazil$ zipinfo '*.zip' | cat
Archive:  jc1.zip
Zip file size: 4116 bytes, number of entries: 1
-rw-r--r--  2.1 unx    11837 bX defN 21-Dec-08 20:50 jc.1
1 file, 11837 bytes uncompressed, 3966 bytes compressed:  66.5%

Archive:  testzip.zip
Zip file size: 8106 bytes, number of entries: 2
-rw-r--r--  3.0 unx      197 tx defN 21-Aug-03 15:12 round-table.gv
-rw-r--r--  3.0 unx     8342 bx defN 21-Aug-03 15:12 round-table.gv.pdf
2 files, 8539 bytes uncompressed, 7651 bytes compressed:  10.4%

Archive:  micro.zip
Zip file size: 6144 bytes, number of entries: 8
-rw-r--r--  2.1 unx    10688 bX defN 19-Sep-30 16:47 microsimservermac.py
drwxrwxr-x  2.1 unx        0 bx stor 21-Dec-20 14:33 __MACOSX/
-rw-r--r--  2.1 unx      176 bX defN 19-Sep-30 16:47 __MACOSX/._microsimservermac.py
-rw-r--r--  2.1 unx      528 bX defN 19-Aug-27 07:46 Dockerfile
-rw-r--r--  2.1 unx    10538 bX defN 19-Oct-01 13:22 microsimserver.py
-rw-r--r--  2.1 unx      380 bX defN 19-Oct-01 13:22 changelog.txt
-rwxr-xr-x  2.1 unx      263 bX defN 19-Oct-01 12:09 dockerhub.sh
-rw-r--r--  2.1 unx      266 bX defN 19-Oct-01 12:09 __MACOSX/._dockerhub.sh
8 files, 22839 bytes uncompressed, 4908 bytes compressed:  78.5%

3 archives were successfully processed.

3 archives were successfully processed.
```
The first one is probably STDOUT and the second STDERR.

kellyjonbrazil:
I can publish v1.17.5 with this parser on PyPi if everything is working ok for you. It's passing tests and it looks like all OS's I have checked are running the latest `zipinfo` version (3.00) from 2009, so I think we're good.

listuser:
ok, sounds good. thanks

PR #188, opened 2021-12-29
Create key value pairs from a MANIFEST.MF file, to include key multiline value pairs.

listuser:
--metamf         MANIFEST.MF file parser

test data in  tests/fixtures/rhel-8/

cmds_used_to_generate_test_data.txt
MANIFEST.MF
MANIFEST.MF.json
unzip_c_MANIFEST.MF.multi_jar.json
unzip_c_MANIFEST.MF.multi_jar.out
unzip_c_MANIFEST.MF.single_jar.json
unzip_c_MANIFEST.MF.single_jar.out


listuser:
Wasn't sure what to call the parser, so I named it "metamf".
Open to suggestions or keep it as is.

kellyjonbrazil:
Sweet - I’ll take a look. Thanks for your contributions!

kellyjonbrazil:
Looks pretty good! Just spitballing... maybe call it `java-manifest`? 

At first I was thinking the multi-line fields should be arrays (split on `;`), but would need to check deeper in the spec to ensure we know which fields could be multi-line/multi-value so we can define them in the schema. May not be worth it.

I was also thinking - maybe just ignore the `inflating: META-INF/MANIFEST.MF` line. Not sure it adds any value to the resulting JSON.

listuser:
Additional suggestions for parser names.

j-manifest

kv-manifest

arch-manifest

Though using "java" or "jar" in the parser name would be helpful to others in determining what the parser does, I chose to stay away from using the words "java" or "jar".

RE: "At first I was thinking the multi-line fields should be arrays (split on ;), but would need to check deeper in the spec to ensure we know which fields could be multi-line/multi-value so we can define them in the schema. May not be worth it." [KB]
I could not find a value in parsing the multi-line fields beyond what I did. Those fields look like class names with version info. Though I don't think they are split on ";".

RE: "I was also thinking - maybe just ignore the inflating: META-INF/MANIFEST.MF line. Not sure it adds any value to the resulting JSON." [KB]
I agree and will update the parser to pop it.

kellyjonbrazil:
Why not `jar-manifest`, since that is the file-type being parsed? Can this parser work on other types of manifest files, too?

jc/parsers/metamf.py
```diff
@@ -0,0 +1,228 @@
...
+import re
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '0.01'
+    description = 'MANIFEST.MF file parser'
+    author = 'Matt J'
+    author_email = 'https://github.com/listuser'
+    compatible = ['linux']
```
kellyjonbrazil, line 93: We can add all architectures here since this is a filetype parser: ['linux', 'darwin', 'cygwin', 'win32', 'aix', 'freebsd']

jc/parsers/metamf.py
```diff
@@ -0,0 +1,228 @@
...
+        "Archive": "apache-log4j-2.16.0-bin/log4j-spring-cloud-config-client-2.16.0.jar",
+        "Manifest_Version": "1.0",
+        "Bundle_License": "https://www.apache.org/licenses/LICENSE-2.0.txt",
+        ...
+      }
+      ...
+    ]
+"""
+import jc.utils
+import jc.parsers.universal
```
kellyjonbrazil, line 83: I don't think any universal parsers are being used, so we can delete this import

jc/parsers/metamf.py
```diff
@@ -0,0 +1,228 @@
...
+
+            # pop key multiline value pair lines
+            if plines:
+                for p in reversed(plines):
+                    archive_item.pop(p)
+
+            # all other key value pairs
+            for i, line in enumerate(archive_item):
+                k, v = line.split(":", maxsplit=1)
+                v = v.rstrip()
```
kellyjonbrazil, line 215: I think just doing a single `v = v.strip()` will do both left and right strip in a single line.

jc/parsers/metamf.py
```diff
@@ -0,0 +1,228 @@
+"""jc - JSON CLI output utility `metamf` command output parser
+
+Options supported:
+- none
+
+Create key value pairs from a MANIFEST.MF file, to include key multiline value pairs.
```
kellyjonbrazil, line 6: Don't need these lines in the docstring for this parser. Since it is a filetype parser, there are no command options to support. The intent of the parser should be clear by the name and metadata.

listuser:
jar-manifest it is.

Thanks for the feedback.
Will make the changes, rename the parser, and update the test outputs as part of my next commit.

kellyjonbrazil:
Could you also include the multi archive test output? I don’t see it anymore in the latest commit.

kellyjonbrazil:
Will need to use `jar_manifest.py` as the module name. The command line argument will still use the dash, though, when the parser is defined in `cli.py`.

kellyjonbrazil:
Nice - thanks for the updates. I'll pull this in and get it in the next release - maybe by today or later this weekend.

Thanks again!

listuser:
sounds good. thank you

PR #190, opened 2022-01-02
fix doubleqoute in csv

shaikustin:
csv files with fields that have doubleqoutes escaping in the form of "" instead of \" were not parsed correctly.
this happened when double qoute appared after sniff (although changing the sniff to the whole file did not work).
#191


kellyjonbrazil:
Thanks, looks good! I'm updating the streaming parser so their behaviors match.

PR #204, opened 2022-02-03
New parser: xrandr

lyterk:
Tests passing, working as intended in plugin directory

It doesn't seem to work from the jc.parsers directory, but I couldn't figure out a reason from reading cli.py.

I have separate tests for types of line (screen, device, mode), and the regexes that identify those lines.


kellyjonbrazil:
Nice! Thank you for the contribution!

I’ll take a closer look in the next few days. Real quick - what are the files under the templates folder? I use that folder for jinja template files that get processed for documentation. 

Also, sorry for the confusion - I recently moved the parser list from cli.py to lib.py so that’s where you can add the parser so you don’t need it to be in the plug-ins directory. 

Thanks again!

lyterk:
Np, sorry it took so long to follow up.

Ah I see, I was using those for testing, I can fix that if you like.

I think I might need to fix up documentation for my parser, I mostly just made the code work and tests pass, didn't look back at your doc format yet.

kellyjonbrazil:
Also noticed you are using the new | operand instead of Union for type annotations. We’ll need to use the old-style (Union) for compatibility with older Python versions. (We officially support and test >=v3.7)

kellyjonbrazil:
```
ERROR: test_xrandr (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_xrandr
  File "/home/runner/work/jc/jc/jc/parsers/xrandr.py", line 102, in <module>
    from typing import TypedDict
ImportError: cannot import name 'TypedDict' from 'typing' (/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/typing.py)
```
Also looks like we cannot use `TypedDict` until Python 3.7 is deprecated.

kellyjonbrazil:
Nice - tests passed! I'll need a few more days to take a closer look and play around with it. Thanks for the updates!

kellyjonbrazil:
Looks good! Just a couple nitpicks:

- String-to-int/float conversions are usually done after parsing in the _process() function, but it looks like you are doing all of the conversions in parse(). Not a big deal since it might be tedious to do the conversions with the dictionary depth. You could just remove _process() if it's not needed. This just means there will be no `jc -r` functionality.

- Please change the test filenames for `simple_xrandr.xxx` to `xrandr_simple.xxx` so the files are grouped together by name.

- It would be nice if no data is sent to the parser (e.g. `echo | jc --xrandr`) it would simply return `{}` instead of `{"screens":[],"unassociated_devices":[]}`. I have a `nodata` test for this in the other parsers.

Other than these and the comments above, looks like we are good to go. No worries on the docs - I can fix those up.

Thanks!

lyterk:
Oh damn that's silly of me. Feel free to merge, I don't think I have any more changes.

PR #223, opened 2022-04-10
Fix/support rotated devices

anekos:
Support rotated devices and avoid entering an infinite loop.


kellyjonbrazil:
Thanks for the parser enhancement! I’ll take a closer look at this for the next release. So far, looks clean!

PR #225, opened 2022-04-12
Update README

adrianteri:
Specify that the Arch package is in the AUR and NOT in the standard Archlinux repositories.


kellyjonbrazil:
Thanks for the update!

PR #240, opened 2022-05-06
Update README.md

fiskhest:
Looks like the arch linux package was moved out of AUR and into the official repositories recently: https://archlinux.org/packages/community/any/jc/


kellyjonbrazil:
Nice - thanks!

fiskhest:
Thank you! :pray:

PR #284, opened 2022-09-14
id: support space in names

pbrezina:
id parser did not work correctly if space is present in user or group
name.


kellyjonbrazil:
Very nice! Thanks for the parser fix. I'll look it over in more detail, but it looks like I'll be able to merge this soon.

Thanks again!

PR #304, opened 2022-10-16
convert domain, bus, dev, function to numbers

braindevices:
I did not convert the vendor/device id because most programs expect hex value only difference is with 0x or not.

### DIFF
--- a/jc/parsers/lspci.py
+++ b/jc/parsers/lspci.py
@@ -109,15 +109,15 @@ def parse(
                 *domain, bus, dev_fun = slot_info.split(':')
 
                 if domain:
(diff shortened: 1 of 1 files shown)


kellyjonbrazil:
Thanks for the PR! What might make sense is to keep the original hex (string) values and add new int fields so the user can choose which ones they want to use. I usually add these fields in the `process` function so the user can choose whether they want those extra fields or not via `--raw` (cli) or `raw=True` (module).

kellyjonbrazil:
Thanks for the suggestions here. I'm manually adding these concepts in via the `_process` function.

Issue #321, opened 2022-11-11
git log parser crashes with empty author name or email


adamwolf:
I am not super familiar with jc, but I was using it the other day to parse `git log`.  It worked great on most of the repositories, but it choked on one.

I looked into it, and a person had set their author string to ' <someemail@address.com>'.  When jc split the author string into name, email, and then grabbed the second part, it crashed since it only split once, with all that leading whitespace.

I added some tests that seemed to cover the situation, and tweaked the parser to pass.  I have to admit I am not very familiar with jc at all, so I do apologize if I missed something obvious :)

I tried to follow similar style as what was there already, but I welcome any suggestions, and if you just want to make the tweaks yourself, feel free.

I can open a PR, but it seemed like I should run the issue by you first.  The work is at https://github.com/adamwolf/jc/tree/git-log-blank-author

Thanks folks!

kellyjonbrazil:
Hi - thank you for your contribution - very much appreciated! I took a quick look and the changes look pretty nice. Please open a PR and I'll look at it in more detail and make sure all the tests are passing.

Thanks again!

kellyjonbrazil:
Thanks for the PR fix - looks good! Tests are passing and docs are updated. This will be included in the next release (v1.22.3)

kellyjonbrazil:
fixed in v1.22.3

PR #322, opened 2022-11-11
Fix git log parsing with empty name or email

adamwolf:
Sometimes, folks leave their name or email blank in on their git commits.  Previously, a blank name crashed the git log parser.

For issue #321


kellyjonbrazil:
Looks good! I did some testing and I like how it changes the author/email to `None` instead of a blank string. I'll fix the broken tests and bump the parser versions after I merge. Thanks again!

Issue #360, opened 2023-02-07
Modify the xrandr parser to have the reflect value in its own JSON key


tzeikob:
Currenty the xrandr parser includes the reflect value of a device in the rotation key of the schema. I would like to propose a small modification to the xrandr parser, so we can have the `reflect` value in its own key like so:

Before:
```json
{
      "associated_modes": [],
      "is_connected": true,
      "is_primary": false,
      "device_name": "VGA-1",
      "rotation": "inverted X and Y axis",
      "resolution_width": 1440,
      "resolution_height": 900,
      "offset_width": 1366,
      "offset_height": 0,
      "dimension_width": 408,
      "dimension_height": 255
    }
```
After:
```json
{
      "associated_modes": [],
      "is_connected": true,
      "is_primary": false,
      "device_name": "VGA-1",
      "rotation": "inverted",
      "reflect": "X and Y axis",
      "resolution_width": 1440,
      "resolution_height": 900,
      "offset_width": 1366,
      "offset_height": 0,
      "dimension_width": 408,
      "dimension_height": 255
}
```

I already have this done and well tested in double monitor setup.
I can push this in PR immediately.

kellyjonbrazil:
Nice! I’ll take a closer look this week.

kellyjonbrazil:
Thanks for the PR. I have merged this into `dev` and it will be included in the next release.

kellyjonbrazil:
Added in v1.23.0

PR #361, opened 2023-02-07
Move the reflect value in its own key (xrandr) #360

tzeikob:
Currently the reflect value of a device is included in the rotation key as a combination of both rotation and reflection (e.g. normal X axis). With this modification we move the reflection value to its own key in the JSON document like so:

```json
{
  ...
  "rotation": "...",
  "reflection": "...,
  ...
}
```

Fixes: issue #360

### DIFF
--- a/jc/parsers/xrandr.py
+++ b/jc/parsers/xrandr.py
@@ -50,7 +50,8 @@ Schema:
           "offset_height":                     integer,
           "dimension_width":                   integer,
           "dimension_height":                  integer,
-          "rotation":                          string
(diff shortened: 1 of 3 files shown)


kellyjonbrazil:
Looks nice - would you mind adding a test with that field included? Thanks!

tzeikob:
@kellyjonbrazil done, let me known if I missed anything.

kellyjonbrazil:
Looks good, thanks! Merging into `dev`. Thanks for contributing!

Issue #364, opened 2023-02-13
Modify the xrandr parser to extract display model name from EDID


tzeikob:
The xrandr tool has an option named `--properties` which causes xrandr to print more information per output, part of this information is the `EDID` hexadecimal string.

> EDID (Extended Display Identification Data) is an industry standard for AV Source and Display devices which automatically communicates manufacturer specification and data between devices.

So decoding the EDID string we can get more data per display, one of which is the model name (e.g. Asus W193). The problem is that calling xrandr along with `--properties` breaks the current implentation of the parser, as this changes the default xrandr output to:

```sh
xrandr --properties
Screen 0: minimum 320 x 200, current 1366 x 768, maximum 8192 x 8192
LVDS-1 connected primary 1366x768+0+0 (normal left inverted right x axis y axis) 344mm x 194mm
        EDID: 
                00ffffffffffff004ca3523100000000
                0014010380221378eac8959e57549226
                0f505400000001010101010101010101
                010101010101381d56d4500016303020
                250058c2100000190000000f00000000
                000000000025d9066a00000000fe0053
                414d53554e470a204ca34154000000fe
                004c544e313536415432343430310018
        scaling mode: Full aspect 
                supported: Full, Center, Full aspect
        link-status: Good 
                supported: Good, Bad
        CONNECTOR_ID: 61 
                supported: 61
        non-desktop: 0 
                range: (0, 1)
   1366x768      60.00*+
   1280x720      60.00    59.99    59.86    59.74  
   1024x768      60.04    60.00  
   960x720       60.00 
   ...
VGA-1 disconnected (normal left inverted right x axis y axis)
        link-status: Good 
                supported: Good, Bad
        CONNECTOR_ID: 64 
                supported: 64
        non-desktop: 0 
                range: (0, 1)
...
``` 

My suggestion is to modify the parser so we can make it hybrid and when xrandr is called with the option `--properties` the parser should try to decode and extract the model name from the EDID string, otherwise should be **backward compatible** and work the same way as it does with outputs from regular xrandr calls (no `--properties`).

So the command `xrandr --properties | jc --xrandr` should give the following JSON document:

```json
{
  "screens": [
    {
      "screen_number": 0,
      "minimum_width": 320,
      "minimum_height": 200,
      "current_width": 1366,
      "current_height": 768,
      "maximum_width": 8192,
      "maximum_height": 8192,
      "associated_device": {
        "associated_modes": [],
        "is_connected": true,
[...]

PR #365, opened 2023-02-17
Modify the xrandr parser to extract display model info from EDID #364

tzeikob:


kellyjonbrazil:
Thanks for submitting! I’ll take a closer look in the next few days.

kellyjonbrazil:
Looks like tests are failing on older versions of python (3.7, 3.6) due to some modern type annotations used. Unfortunately we need to be backward compatible to python 3.6 until Centos7 is no longer supported. Could you modify/remove the offending type annotations?

jc/parsers/xrandr.py
```diff
@@ -294,15 +375,67 @@ def _parse_device(next_lines: List[str], quiet: bool = False) -> Optional[Device
...
+#      010101010101381d56d4500016303020
+#      250058c2100000190000000f00000000
+#      000000000025d9066a00000000fe0053
+#      414d53554e470a204ca34154000000fe
+#      004c544e313536415432343430310018
+_edid_head_pattern = r"\s*EDID:\s*"
+_edid_line_pattern = r"\s*(?P<edid_line>[0-9a-fA-F]{32})\s*"
+
+
+def _parse_model(next_lines: List[str], quiet: bool = False) -> Optional[Model]:
```
kellyjonbrazil, line 409: It looks like the type annotation here (specifically referring to `Model`) is breaking python versions 3.6 and 3.7. Not sure why, but maybe we just modify/remove the type annotation here and see if that will allow the tests to pass? Type annotations seem to be a bit broken in this parser, overall.

tzeikob: Sure, we can do that.

The only files the xrandr parser needs, in order to decode the EDID data, is the Edid and EdidHelper classes. I think we can remove any other file entirely and keep those classes having them modified a bit by removing any @staticmethod, @classmethod annotations.

Let me know if there is a way to run the tests which are failing for centos to save some from back and forth time. I suppose the `runtests.sh` script does not include those tests, right? I ran this script and all test are passing.

kellyjonbrazil: All the tests run with `./runtests.sh`, but github actions just runs the same tests several times with different Python versions and operating systems. The way to do it on your local machine is by using something like [pyenv](https://github.com/pyenv/pyenv) so you can change to an older Python version and run `./runtests.sh` again.

tzeikob:
@kellyjonbrazil I think I've found the root of the cause.

I was actually missed entirely the part where typed dictionaries must be declared as generic ones in case the TypedDict is not supported, as it turns out for older versions of python.

kellyjonbrazil:
Looks great - I'll go ahead and merge. Thank you for the contribution!

PR #378, opened 2023-03-18
Make umask optional in /proc/PID/status

jwilk:
Zombie processes don't have the umask field available.


kellyjonbrazil:
Thanks - this will make it in the next release!

PR #384, opened 2023-03-31
Fix iwconfig re_interface regex

six3six:
Small fix of re_interface in iwconfig.py.
When the network SSID contains a "-" character, the network interface was ignored.


vrince:
Was about to fix it! encountered the same issue this morning ... Will drop my MR to keep this one.

vrince:
Actually, I have no idea how "weird" ESSID could be, so I relax way more with the regex in my MR maybe it's gonna be a little bit more "future" proof? #386 What so you think?

kellyjonbrazil:
Thanks for the fix! This is expectedly causing tests to fail, but I can merge this and fix them up as well as add a test case for SSIDs with a `-` character.

kellyjonbrazil:
Oh sorry that was another PR causing tests to fail. It looks like the other PR was stripping fields, which is probably a good idea. I'll merge this and also add that functionality.

PR #386, opened 2023-03-31
iwconfig : relax regex to match interface

vrince:
I encountered real-life examples where the interface was not matching so the parser return nothing.
I think it could be way more relaxed and remain safe.


vrince:
PR already available to fix when `-` is in the ESSID.

six3six:
This PR should most certainly work, but in the future, if you want to be fully RFC compliant it might be a good idea to build a regex that get all printable ASCII characters except : ?, ", $, [, \, ], and +.

I found a good (and short) doc from Cisco that explains the RFC : https://www.cisco.com/assets/sol/sb/WAP321_Emulators/WAP321_Emulator_v1.0.0.3/help/Wireless05.html

PR #402, opened 2023-05-01
Fix for extra opts

gaugstein:
Fix for missing header information and the resulting different field lengths


kellyjonbrazil:
I'll close this PR as it seems to be a duplicate of the previous one. As you add commits to your branch they will automatically make it into the original PR.

PR #408, opened 2023-05-19
Netstat parser: windows support

YeahItsMeAgain:
Hi :)

Supports `netstat -anbo`
I didn't parse `netstat -r` because it's basically the same as `route print` (so a new parser should be created).

Update:
I created a new parser for `route print`(https://github.com/kellyjonbrazil/jc/pull/410) but the output seems different from the output of the netstat parser, so i decided to keep out `netstat -r`, it's up to you to decide if to keep it that way :)


kellyjonbrazil:
Very nice! Thanks for your contributions! I'll start working on the next release soon.

PR #413, opened 2023-05-26
New Parser: lsattr

YeahItsMeAgain:


jc/parsers/lsattr.py
```diff
@@ -0,0 +1,157 @@
...
+
+ERROR_PREFIX = "lsattr:"
+
+# https://github.com/mirror/busybox/blob/2d4a3d9e6c1493a9520b907e07a41aca90cdfd94/e2fsprogs/e2fs_lib.c#L40
+# https://github.com/landley/toybox/blob/f1682dc79fd75f64042b5438918fe5a507977e1c/toys/other/lsattr.c#L97
+ATTRIBUTES = {
+    "B": "Compressed_File",
+    "Z": "Compressed_Dirty_File",
+    "X": "Compression_Raw_Access",
+    "s": "Secure_Deletion",
```
kellyjonbrazil, line 86: Could we lower-case the dict values here? Makes it nicer to use via command-line so the JSON key-names will be all lower with underscores.

jc/parsers/lsattr.py
```diff
@@ -0,0 +1,157 @@
...
+
+import jc.utils
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '1.0'
+    description = '`lsattr` command parser'
+    author = 'Kelly Brazil'
+    author_email = 'kellyjonbrazil@gmail.com'
```
kellyjonbrazil, line 69: Could you add your info here?

jc/parsers/lsattr.py
```diff
@@ -0,0 +1,157 @@
...
+Information from https://github.com/mirror/busybox/blob/2d4a3d9e6c1493a9520b907e07a41aca90cdfd94/e2fsprogs/e2fs_lib.c#L40
+used to define field names
+
+    [
+      {
+        "file":                           string,
+        "Compressed_File":                Optional[boolean],
+        "Compressed_Dirty_File":          Optional[boolean],
+        "Compression_Raw_Access":         Optional[boolean],
+        "Secure_Deletion":                Optional[boolean],
```
kellyjonbrazil, line 27: This would need to be updated after changing the key names to lower case.

kellyjonbrazil (changes requested):
Thank you for the contribution!

kellyjonbrazil:
Looks great, thanks! I'll merge and get this into the next release.

PR #415, opened 2023-06-02
New parser: srt

### LINKED ISSUE #411: [new parser request] .srt subtitle file
https://en.wikipedia.org/wiki/SubRip

Presumably would be trivial to implement: remove lines containing the subtitle track numbers and timestamps, merge split lines.

YeahItsMeAgain:
Fixes #411.

I'm not sure how to handle special formatting (https://docs.fileformat.com/video/srt/#formatting-of-srt-files), but i guess this is a start.


kellyjonbrazil:
Very cool - you rock! I'll take a closer look but so far looks great.

jc/parsers/srt.py
```diff
@@ -0,0 +1,204 @@
...
+    for line in lines:
+        line = line.strip()
+        if line.isdigit():  # Reached a new subtitle.
+            if subtitle:
+                raw_output.append(subtitle)
+                subtitle = {}
+            subtitle['index'] = line
+            continue
+
+        if '-->' in line:  # Start and end time of the subtitle separated by –> characters.
```
kellyjonbrazil, line 192: I wonder if we should tighten this up a bit? I really can't think of why subtitles would have `-->` in them, so this is probably not a real issue, but I was thinking maybe we should use a regex that matches the `00:02:16,612 --> 00:02:19,376` pattern instead. Something like:

```
^\d\d:\d\d:\d\d,\d+ --> \d\d:\d\d:\d\d,\d+$
```

kellyjonbrazil (commented):
Looks good - just a minor thought on the start/end time matching.

jc/parsers/srt.py
```diff
@@ -0,0 +1,204 @@
...
+
+    raw_output = []
+    if not jc.utils.has_data(data):
+        return raw_output
+
+    subtitle = {}
+    lines = list(filter(None, data.splitlines()))
+    for line in lines:
+        line = line.strip()
+        if line.isdigit():  # Reached a new subtitle.
```
kellyjonbrazil, line 185: Clever - I like it but wonder if we need to tighten this one a bit more. Maybe accounting for a newline-only before the digit (except maybe the first one)? Just in case there is a subtitle with only a single number. Just a thought.

YeahItsMeAgain:
HI :)

so yeah, my previous attempt was naive. It didn't support the edge cases you mentioned.

I researched what other parsers do and found this great library (with an MIT license 😅 https://github.com/cdown/srt ) & copied the relevant parts from there.
let me know if there are more changes to be made 🔥

btw, noticed you don't squash pull request merged, why is that?

kellyjonbrazil:
Changes look great, thanks!

Regarding the squashing of Pull Requests - not sure, never really thought about it. Should I be doing that? ("Squash and Merge" instead of "Create a Merge Commit"?)

YeahItsMeAgain:
The commit history of a pull request is pretty much useless so condensing it to 1 commit usually looks nicer on the git history :)

PR #422, opened 2023-06-19
Last -x

YeahItsMeAgain:
Hi :)
Added support for `-x` option, specifically `runlevel` (it has spaces so entry.split() gets a bit messed up).

Refactored the parser a bit to make it easier to add changes, let me know if there are more changes you want me to make.


kellyjonbrazil:
Looks great, thanks! I'll get this in the next release.

PR #423, opened 2023-06-20
last -x shutdown fix

YeahItsMeAgain:
Hi :)

While testing my changes on another server ([!422](https://github.com/kellyjonbrazil/jc/pull/422)) i noticed i missed the "shutdown" entry, my bad.

I think there is another entry("new time"\\"old time") (https://github.com/util-linux/util-linux/blob/33e347832e77e821c62cfbc48eccff178f9ea6b5/login-utils/last.c#L782) but i can't replicate it on my machines :/


kellyjonbrazil:
Sweet, thanks!

PR #429, opened 2023-06-22
Completed Ip_route parser

Julian5555:
Completed the ip_route parser. Let me know if any formatting or changes need to be made.

### DIFF
--- a/jc/parsers/ip_route.py
+++ b/jc/parsers/ip_route.py
@@ -1,0 +1,142 @@
+"""jc - JSON Convert `ip route` command output parser
+
+Usage (cli):
+
+    $ ip route | jc --ip_route
+
+or
+
+    $ jc ip_route
+
+Usage (module):
+
+    import jc
+    result = jc.parse('ip route', ip_route_command_output)
+
+Schema:
+
+    [
+      {
+        "ip":        string,
+        "via":       string,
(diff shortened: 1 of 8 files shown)


jc/parsers/ip_route.py
```diff
@@ -0,0 +1,142 @@
+"""jc - JSON Convert `ip route` command output parser
+
+Usage (cli):
+
+    $ ip route | jc --ip_route
+
+or
+
+    $ jc ip_route
```
kellyjonbrazil, line 9: The `jc` argument will be `ip-route` with a dash here.

kellyjonbrazil (changes requested):
Nice work - thanks for the contribution!

Julian5555 (commented):
I reformatted the requested changes and changed the metric key into an int for non raw data while leaving r data as a dictionary of strings as per you're request. Also used if's instead of the previous match cases so that the parser should now be backwards compatible. If I missed anything else, please let me know. Thanks for the review.

kellyjonbrazil (approved):

kellyjonbrazil:
Thanks for the updates! I'll merge and do a little cleanup.

Julian5555:
Looks great, I appreciate the formatting help and you taking the time to review this.

PR #430, opened 2023-06-22
Added find parser and tests for Centos 7.7 and Ubuntu 18.04

solomonleang:
I added a parser for the find command in Linux and a some tests for Centos 7.7 and Ubuntu 18.04. Please let me know if there are any changes you would like to be made or fixed!

### DIFF
--- a/jc/parsers/find.py
+++ b/jc/parsers/find.py
@@ -1,0 +1,121 @@
+"""jc - JSON Convert `find` command output parser
+
+Usage (cli):
+
+    $ find | jc --find
+
+Usage (module):
+
+    import jc
+    result = jc.parse('find', find_command_output)
+
+Schema:
+
+    [
+      {
+        "directory path":           string,
+        "file or empty directory:   string
+      }
+    ]
+
+Examples:
+
+    $ find | jc --find -p
+    [
+        {
+          "directory path": "./directory"
+          "file or empty directory": "filename"
+        },
+        {
+          "directory path": "./anotherdirectory"
+          "file or empty directory": "anotherfile"
+        },
+        ...
+    ]
+
+    $ find | jc --find -p -r
+    {
+      "./templates/readme_template",
+      "./templates/manpage_template",
+      "./.github/workflows/pythonapp.yml",
+      ...
+    }
+"""
+import jc.utils
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '1.0'
+    description = '`find` command parser'
+    author = 'Solomon Leang'
+    author_email = 'solomonleang@gmail.com'
+    compatible = ['linux']
+    tags = ['command']
+
+
+__version__ = info.version
+
+
+def _process(proc_data):
+    """
+    Final processing to conform to the schema.
+
+    Parameters:
+
+        proc_data:   (Dictionary) raw structured data to process
+
+    Returns:
+
+        List of Dictionaries. Structured data to conform to the schema.
+    """
+
+    processed = []
+    for i in proc_data:
+        try:
+            temp = str(i).rindex("/")
+        except ValueError:
+            pass
+        proc_line = {
+            'directory path': i[:temp + 1],
+            'file or empty directory': i[temp + 1:]
+        }
+        
+        processed.append(proc_line)
+    return processed
+    
+
+
+def parse(data, raw=False, quiet=False):
+    """
+    Main text parsing function
+
+    Parameters:
+
+        data:        (string)  text data to parse
+        raw:         (boolean) unprocessed output if True
+        quiet:       (boolean) suppress warning messages if True
+
+    Returns:
+
+        Dictionary of raw structured data or
+        List of Dictionaries of processed structured data
+    """
+    jc.utils.compatibility(__name__, info.compatible, quiet)
+    jc.utils.input_type_check(data)
+
+    raw_output = []
+    
+    if jc.utils.has_data(data):
+        previous = ""
+        new_line = data.split("\n")
+        for index in new_line:
+            if (index[:len(previous)] != previous):
+                raw_output.append(previous)
+            previous = index
+        raw_output.append(previous)
+
+    if raw:
+        return raw_output
+    else:
+        return _process(raw_output)
--- a/jc/lib.py
+++ b/jc/lib.py
@@ -43,6 +43,7 @@ parsers: List[str] = [
     'email-address',
     'env',
     'file',
+    'find',
     'findmnt',
     'finger',
     'free',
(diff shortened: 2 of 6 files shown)


jc/parsers/find.py
```diff
@@ -0,0 +1,121 @@
...
+    result = jc.parse('find', find_command_output)
+
+Schema:
+
+    [
+      {
+        "directory path":           string,
+        "file or empty directory:   string
+      }
+    ]
```
kellyjonbrazil, line 19: Can we shorten the key names and change spaces to underscore? Something like: `path` and `file`. Or maybe something like `node` makes more sense instead of `file` since it can represent an empty directory?

jc/parsers/find.py
```diff
@@ -0,0 +1,121 @@
...
+        ...
+    ]
+
+    $ find | jc --find -p -r
+    {
+      "./templates/readme_template",
+      "./templates/manpage_template",
+      "./.github/workflows/pythonapp.yml",
+      ...
+    }
```
kellyjonbrazil, line 42: Think this should be a list (square brace instead of curly)

jc/parsers/find.py
```diff
@@ -0,0 +1,121 @@
...
+
+        List of Dictionaries. Structured data to conform to the schema.
+    """
+
+    processed = []
+    for i in proc_data:
+        try:
+            temp = str(i).rindex("/")
+        except ValueError:
+            pass
```
kellyjonbrazil, line 78: Think you should set temp above the try/except or within the except so it is assigned even if there is a ValueError. Otherwise you'll get a name error when trying to use temp below.

Maybe an easier way of assigning the path/file would be something like:
```
path, file = i.rsplit('/', maxsplit=1)
```
Then you can just use `path` and `file` as the values below instead of slicing.

jc/parsers/find.py
```diff
@@ -0,0 +1,121 @@
...
+        List of Dictionaries of processed structured data
+    """
+    jc.utils.compatibility(__name__, info.compatible, quiet)
+    jc.utils.input_type_check(data)
+
+    raw_output = []
+    
+    if jc.utils.has_data(data):
+        previous = ""
+        new_line = data.split("\n")
```
kellyjonbrazil, line 111: Could we use `data.splitlines()` here instead?

jc/parsers/find.py
```diff
@@ -0,0 +1,121 @@
...
+        for index in new_line:
+            if (index[:len(previous)] != previous):
+                raw_output.append(previous)
+            previous = index
+        raw_output.append(previous)
```
kellyjonbrazil, line 116: Not sure I understand what this is doing. Is this removing duplicates?

kellyjonbrazil (changes requested):
Thanks for the parser contribution! Could you take a look at these items? Also, Could you add the test file?

jc/parsers/find.py
solomonleang: Essentially what it is doing is checking to see if the previous path is contained in the current index (path). This would ensure every directory is accounted for and all files (or empty directories) would get added into the list of raw_output. The last statement there ensures the last file gets added to the list. Let me know if this makes more sense!

solomonleang:
Thanks for the review, I will add the test file and fix those items soon!

jc/parsers/find.py
kellyjonbrazil: I don't understand why, though. If the `find` output is just a newline terminated list of paths like this:
```
.
./
./jc
./jc/tests
./jc/tests/test_find.py
./jc/tests/test_history.py
./jc/tests/test_hosts.py
./jc/anotherdirectory
```
Wouldn't just running it through `splitlines()` to turn it into a list do the same thing? Or are you checking for paths with newline characters in them or something? I'm probably missing something simple. :)

solomonleang: Ah I see what you mean. I am removing the lines where it is only the directory being listed (and that the directory is not empty).
So in the example you provided, it would skip these:
```
.
./
./jc
./jc/tests
```
because they aren't files and the directory is not empty.

Thats why the output would only be something like below (also assuming anotherdirectory is empty in this case!)
```
[{"directory path": "./jc/tests/","file or empty directory": "test_find.py"},
{"directory path": "./jc/tests/","file or empty directory": "test_history.py"},
{"directory path": "./jc/tests/","file or empty directory": "test_hosts.py"},
{"directory path": "./jc/","file or empty directory": "anotherdirectory"}]
``` 

However, I can include these 4 lines instead and change the schema so that it is just `path` and `node` where `node` can be a nonempty directory as well; or, would skipping these lines still be acceptable?

kellyjonbrazil: Interesting. I see how skipping those lines makes sense, but on the other hand, they are part of the find output so it seems we should just leave them in. The downstream application can decide to ignore those lines if they want with their own logic.

solomonleang:
I made the changes that you requested! Thanks for the feedback.
I did some researching and saw that find could also return a Permission denied error so I included that in my schema as well.

jc/parsers/find.py
```diff
@@ -0,0 +1,126 @@
...
+
+        List of Dictionaries. Structured data to conform to the schema.
+    """
+
+    processed = []
+    for index in proc_data:
+        path, node, error = "", "", ""
+        if (index == "."):
+            node = "."
+        elif (index[:4] == "find"):
```
kellyjonbrazil, line 81: Could use `elif index.startswith('find: '):` here

jc/parsers/find.py
```diff
@@ -0,0 +1,126 @@
...
+            try:
+                path, node = index.rsplit('/', maxsplit=1)
+            except ValueError:
+                pass
+        
+        proc_line = {
+            'path': path,
+            'node': node,
+            'error': error
+        }
```
kellyjonbrazil, line 93: We could do something like this so error field only shows up when needed:
```
proc_line = {
    'path': path if path else None,
    'node': node if node else None
}

if error:
    proc_line.update(
        {'error': error}
    )
```

That will make the values `null` instead of blank strings if they are empty which seems cleaner to me.

kellyjonbrazil (changes requested):
Looking great! just a couple more thoughts and then we can merge.

jc/parsers/find.py
solomonleang: Yep! Just updated for this. I agree it does look cleaner.

kellyjonbrazil (approved):

solomonleang:
No other changes and ready to merge! Thanks for all the feedback provided!

kellyjonbrazil:
My bad, this was supposed to go into `dev` not `master`. Had to revert. I'll try to figure this out.

kellyjonbrazil:
Can you re-open a PR, this time from the `dev` branch? I think I have things back to normal now. Thanks!

solomonleang:
Yep just made another PR from the dev branch

PR #450, opened 2023-09-08
Add ISC 'host' support

pettai:
Add ISC 'host' command support


kellyjonbrazil:
Nice - thanks for the parser contribution! Could you add the integer conversion in the `_process()` function? It would also be great to add some tests.

pettai:
> Nice - thanks for the parser contribution! Could you add the integer conversion in the `_process()` function? It would also be great to add some tests.

Ok, I'll try to do the integer part.
About tests, do you have any pointers? if there was in foo.py, I must have missed them...

kellyjonbrazil:
Thanks for adding the integer conversions! Here is a template you can use for the tests:
https://github.com/kellyjonbrazil/jc/blob/master/tests/templates/_test_foo.py

Go ahead and add your test file to the `tests` folder. You can check out https://github.com/kellyjonbrazil/jc/blob/master/tests/test_certbot.py as an example. Put your output file(s) in the appropriate folder under `tests/fixtures`.  Let me know if you need any assistance. Happy to help!

Issue #453, opened 2023-09-12
xrandr parser screen association issue


marcin-koziol:
It seems that in this situation, both outputs should be in {'screens': [{'screen_number': 0, [...] associated_devices array

```
Screen 0: minimum 320 x 200, current 1920 x 1080, maximum 16384 x 16384
DP-1 disconnected primary (normal left inverted right x axis y axis)
DP-2 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 521mm x 293mm
   1920x1080     60.00*+  59.94    60.00  
   1680x1050     60.00    59.88  
   1400x1050     60.00  
   1600x900      60.00  
   1280x1024     75.02    60.02    60.00  
```
meanwhile DP-2 is in unassociated_devices

kellyjonbrazil:
@lyterk could you take a look at this?

lyterk:
Sure, taking a look

lyterk:
@kellyjonbrazil This should be ready to review, lemme know if you want me to bust out the commits more discretely

kellyjonbrazil:
Thanks @lyterk! I'll take a look.

kellyjonbrazil:
Added in v1.23.5

PR #455, opened 2023-09-16
[xrandr] Fix 453 devices issue

lyterk:
Addresses #453 

* Fix: 'devices' was originally not a list, just assigned each time it
was parsed. Made that a list and appended to it.
* Removed distinction between unassociated/associated devices
* Added test for @marcin-koziol's problem
* Put tests into separate test methods


kellyjonbrazil:
Thanks! I'll merge this in this weekend.

Issue #457, opened 2023-09-21
x509_cert parser crashes for certs with superfluous bits in bit string fields


sumezulike:
Hey Kelly, 

I stumbled upon a new edge case that I think should be handled gracefully.

The code for `BitString` seems to assume that bit strings won't contain more bits than defined by the respective standard. I don't know why, but some do.

https://github.com/kellyjonbrazil/jc/blob/4cd721be8595db52b620cc26cd455d95bf56b85b/jc/parsers/asn1crypto/core.py#L2333-L2338

This is one of the certificates I found on Censys which has 16 bits set for the key usage extension. 
[99580524040c341ed12596b5b4d3e09815f769ab70d6266b7e1a31f595c5dffc.zip](https://github.com/kellyjonbrazil/jc/files/12684992/99580524040c341ed12596b5b4d3e09815f769ab70d6266b7e1a31f595c5dffc.zip)

`_map` only contains the keys 0 through 9, as specified by [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3), so line `2337` adds the bits 10 to 15 to the native representation (a set of strings) as integers. That causes a `TypeError: '<' not supported between instances of 'int' and 'str'` when `_fix_objects` tries to sort them.
https://github.com/kellyjonbrazil/jc/blob/3249a017aecc18eeca5a4d36d749744c5b97f06e/jc/parsers/x509_cert.py#L511-L513

OpenSSL just ignores the undefined bits, which I think is the most sensible cause of action.

kellyjonbrazil:
Added in v1.23.5

PR #458, opened 2023-09-21
Adds graceful handling of superfluous bits in bit strings

sumezulike:
Suggested fix for #457

This is the simplest fix I could think of. 
It couldn't see any reason to add the index to the native representation, so that just doesn't happen anymore.
One might need to check the type or so if something else does needs the indices.


sumezulike:
Very sorry about the other commit being added here. I'm not very experienced with git and pull requests. I'll try to remove it.

PR #459, opened 2023-09-21
Adds graceful handling of superfluous bits in bit strings

sumezulike:
Suggested fix for #457

This is the simplest fix I could think of.
It couldn't see any reason to add the index to the native representation, so that just doesn't happen anymore.
One might need to check the type or so if some other parser does needs the indices.


kellyjonbrazil:
Nice - thank you for the fix! I'll merge this in this weekend.

PR #484, opened 2023-11-20
feat(iftop): add iftop-scanning

georgettica:
this is not even an MVP, but I would like it to exist to allow per client json aggregation

also, a future use is a stream response


jc/parsers/iftop.py
```diff
@@ -0,0 +1,397 @@
...
+        "device": None,
+        "ip_address": None,
+        "mac_address": None,
+        "clients": None,
+        "total_send_rate": None,
+        "total_receive_rate": None,
+        "total_send_and_receive_rate": None,
+        "peak_rate": None,
+        "cumulative_rate": None,
+    }
```
kellyjonbrazil, line 192: You can delete this section. This was for backwards compatibility for another parser.

jc/parsers/iftop.py
```diff
@@ -0,0 +1,397 @@
...
+
+    interface_item: Dict = interface_obj.copy()
```
kellyjonbrazil, line 194: This can be a normal empty dict. I was just copying in the other parser so it would keep the default values for backward compatibility.

kellyjonbrazil:
Looking good!

completions/jc_bash_completion.sh
kellyjonbrazil, line 1: You can delete the completion scripts as I auto generate them with the documentation.

requirements.txt
kellyjonbrazil, line 1: We shouldn't introduce any new requirements/dependencies. There may be a way to vendor this in or create our own function that does something similar.

completions/jc_bash_completion.sh
kellyjonbrazil: Or maybe not delete them, but revert them back to original. You should only need to update the parser.py file and the lib.py file for a new parser. I'll handle all the documentation and completion updates when creating the release.

requirements.txt
kellyjonbrazil: It looks like we can pull the `tokenize()` function and the `parse_size()` functions from the library. If you want, you can put them directly in the parser and I'll move them to utils.py later.

kellyjonbrazil: https://github.com/xolox/python-humanfriendly/blob/master/humanfriendly/text.py#L402

https://github.com/xolox/python-humanfriendly/blob/master/humanfriendly/__init__.py#L56-L70

https://github.com/xolox/python-humanfriendly/blob/master/humanfriendly/__init__.py#L198

kellyjonbrazil:
Sweet! I'm on a plane now coming at you from 30k feet and will be traveling a bit this week, so I may be able to merge this later this week or this weekend. Also, I can possibly take your work and turn this into a streaming parser.

jc/parsers/iftop.py
```diff
@@ -0,0 +1,622 @@
+"""jc - JSON Convert `iftop` command output parser
+
+No `iftop` options are supported.
```
kellyjonbrazil, line 3: Are there only certain options that are supported or can we remove this if all options are supported? From the example and tests it looks like at least some options are supported.

Issue #487, opened 2023-11-21
 New parser request: Alpine Linux packages index


don-rumata:
Hi! Can you add a Alpine Linux Packages parser?

https://wiki.alpinelinux.org/wiki/Apk_spec#APKINDEX_Format

Example: `curl --silent http://dl-cdn.alpinelinux.org/alpine/latest-stable/main/x86_64/APKINDEX.tar.gz | zcat`

kellyjonbrazil:
Sure, looks pretty easy.

r-darwish:
So I suppose what's actually supposed to be parsed here is the APKINDEX file inside the tar itself.

@kellyjonbrazil should this parser receive the extracted file as the input, or should it extract & parse it from the tar.gz?

Also, here's a package from the file
```
C:Q1znBl9k+RKgY6gl5Eg3iz73KZbLY=
P:yasm
V:1.3.0-r4
A:x86_64
S:772109
I:1753088
T:A rewrite of NASM to allow for multiple syntax supported (NASM, TASM, GAS, etc.)
U:http://www.tortall.net/projects/yasm/
L:BSD-2-Clause
o:yasm
m:Natanael Copa <ncopa@alpinelinux.org>
t:1681228881
c:84a227baf001b6e0208e3352b294e4d7a40e93de
D:so:libc.musl-x86_64.so.1
p:cmd:vsyasm=1.3.0-r4 cmd:yasm=1.3.0-r4 cmd:ytasm=1.3.0-r4

C:Q1D3O+vigqMNGhFeVW1bT5Z9mZEf8=
P:yasm-dev
V:1.3.0-r4
A:x86_64
S:449076
I:1773568
T:A rewrite of NASM to allow for multiple syntax supported (NASM, TASM, GAS, etc.) (development files)
U:http://www.tortall.net/projects/yasm/
L:BSD-2-Clause
o:yasm
m:Natanael Copa <ncopa@alpinelinux.org>
t:1681228881
c:84a227baf001b6e0208e3352b294e4d7a40e93de
```
I understand most of the field names but not all of them. Can anyone elaborate?

kellyjonbrazil:
This is ready to ship in the next release.

don-rumata:
@kellyjonbrazil, the `apkindex` key is the final version? Or will it be like in https://github.com/kellyjonbrazil/jc/issues/480? I.e. `--apk-packages-index`.

kellyjonbrazil:
[...]

PR #491, opened 2023-11-28
APKINDEX parser (#487)

r-darwish:
#487


kellyjonbrazil:
Thanks for the parser contribution! Looks pretty good - I’ll take a closer look in the next few days.

PR #493, opened 2023-12-01
Remove redundant Python 2 code

hugovk:


kellyjonbrazil:
Hi there - thank you for your contribution!

I was about to say that I'm conflicted about this PR because this is vendored code from another project and part of me just wants to keep from touching it so it doesn't break.

But then I saw who authored this PR, and frankly, I'm honored. 😄 

It looks like all the tests are passing, so I can get this into the next release.

hugovk:
You're welcome :)

> I was about to say that I'm conflicted about this PR because this is vendored code from another project and part of me just wants to keep from touching it so it doesn't break.

Ah, for vendored code, if you're planning on updating it in the future, maybe they've already removed the old code. So we can revert that if you like? Especially if it would make it easier for you to update in the future?

kellyjonbrazil:
I think the PR is fine. There have been a few changes to the vendored code over time anyway, so it's no longer tracking the original library anyway. I was more concerned that someone who knew what they were doing made the updates because some of the code in that library is a mystery to me. These updates all make sense and tests are passing, so I'm glad to get this cleanup merged in.

PR #514, opened 2024-01-04
CONTRIBUTING.md: add setup info / develop mode / pip install

muescha:
add:
- setup python libs
- setup requirements
- setup dev mode
- run test

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,18 @@ Use `--debug` to see error messages (double to see more):
 echo 'abc' | jc --parser-with-error --debug
 ```
 
+## Install `jc` with `pip` from source
+
+Please note: Development mode is preferred for observing instant changes.
+
+```shell
+pip install .
+
+pip freeze | grep jc  # check if jc is installed 
+
+pip uninstall jc
+```
+
 ## Parser Schema Guidelines
 - Try to keep the schema as flat as possible - typically a list of flat dictionaries
 - Keys should be lowercase, contain no special characters, and spaces should be converted to underscores
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ Pull requests are the best way to propose changes to the codebase (we use [Githu
 7. Make sure your code lints.
 8. Issue that pull request!
 
-## Setup Debug Environment
+## Setup Environment
 
 Install requirements:
 
@@ -35,34 +35,26 @@ Install requirements:
 pip install -r requirements.txt
 ```
 
+## Development Mode
+
 Use jc in development mode (changes are direct available):
 
 ```shell
-python3 setup.py develop
+./install.sh
 ```
 
 Uninstall local development:
 
 ```shell
-python3 setup.py develop --uninstall
+./uninstall.sh
 ```
+
+## Debug Messages
 
 Use `--debug` to see error messages (double to see more):
 
 ```shell
 echo 'abc' | jc --parser-with-error --debug
-```
-
-## Install `jc` with `pip` from source
-
-Please note: Development mode is preferred for observing instant changes.
-
-```shell
-pip install .
-
-pip freeze | grep jc  # check if jc is installed 
-
-pip uninstall jc
 ```
 
 ## Parser Schema Guidelines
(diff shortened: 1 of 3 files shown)


muescha:
Any suggestions to improve the wording or correct typos are appreciated.

CONTRIBUTING.md
```diff
@@ -27,6 +27,44 @@ Pull requests are the best way to propose changes to the codebase (we use [Githu
...
+```shell
+echo 'abc' | jc --parser-with-error --debug
+```
+
+## Install `jc` with `pip` from source
+
+Please note: Development mode is preferred for observing instant changes.
+
+```shell
+pip install .
```
kellyjonbrazil, line 61: Should we do something like:
```
pip3 install --upgrade --user -e .
```
This is what `install.sh` does. It basically puts it in developer mode for instant changes and only for the current user.

muescha: ah - I was not aware of this `-e` option :)

I would apply your suggestion to the text

muescha: the pip install not works on my mac :(
I have only success with: `python3 setup.py develop`

```shell
$ jc --version
jc version:  1.24.1
python interpreter version:  3.11.6
python path:  /opt/homebrew/opt/python@3.11/bin/python3.11

https://github.com/kellyjonbrazil/jc
© 2019-2024 Kelly Brazil

$ ./uninstall.sh 
Found existing installation: jc 1.24.1
Uninstalling jc-1.24.1:
  Would remove:
    /opt/homebrew/bin/jc
    /opt/homebrew/lib/python3.11/site-packages/jc.egg-link
Proceed (Y/n)? y
  Successfully uninstalled jc-1.24.1
$ jc --version 
zsh: command not found: jc
$ ./install.sh 
Obtaining file:///Users/muescha/Work/github.com/muescha/jc
  Preparing metadata (setup.py) ... done
Requirement already satisfied: ruamel.yaml>=0.15.0 in /opt/homebrew/lib/python3.11/site-packages (from jc==1.24.1) (0.18.5)
Requirement already satisfied: xmltodict>=0.12.0 in /opt/homebrew/lib/python3.11/site-packages (from jc==1.24.1) (0.13.0)
Requirement already satisfied: Pygments>=2.3.0 in /opt/homebrew/lib/python3.11/site-packages (from jc==1.24.1) (2.17.2)
Requirement already satisfied: ruamel.yaml.clib>=0.2.7 in /opt/homebrew/lib/python3.11/site-packages (from ruamel.yaml>=0.15.0->jc==1.24.1) (0.2.8)
Installing collected packages: jc
  Running setup.py develop for jc
Successfully installed jc-1.24.1
$ jc --version
zsh: command not found: jc

```

muescha: tested some options. Without `--user` I have success:

```shell
pip3 install --upgrade -e .
```

install.sh
```diff
@@ -1,2 +1,4 @@
 #!/bin/bash
-pip3 install --upgrade --user -e .
\ No newline at end of file
+
+# install jc in development mode
+pip3 install --upgrade --editable .
```
muescha, line 4: is it ok to remove the `--user` option?

kellyjonbrazil: Ah, I see - I think I had to add to my path to get the `--user` part to work:
```
export PATH="$HOME/.local/bin:$PATH"
```
Ultimately, I think it would be better to describe a more general/generic approach:
- Use something like `pyenv` and/or `venv` to install `jc` into a virtual environment. That way it doesn't mess with your system python or other libraries.
- Install the requirements into the virtual environment
- Then use the `install.sh` and `unistall.sh` files you created. I think with the above we can remove the `--user` flag

muescha: i have python3 installed long time ago with `brew`. 
on new installs I would use `asdf` or `mise`(formerly known as `rtx`)

I get it working in IntelliJ IDEA

So I don't know how to do the (1) and (2) :/

kellyjonbrazil: Now that I look at it again, the reason for the `--user` flag is so the console-scripts entry point (i.e the `jc` script that gets installed in the path to run the CLI) does not conflict with other installations of `jc`, including installs via a package manager like `brew`.

kellyjonbrazil:
Thank you for this PR! I won't merge this, but I have taken many of your suggestions and have added them to the [CONTRIBUTING.md](https://github.com/kellyjonbrazil/jc/blob/dev/CONTRIBUTING.md) file.

PR #516, opened 2024-01-06
extract to public jc.get_parser(parser_name)

muescha:
extract to public `jc.get_parser(parser_name)`

still now it is only possible to do a:

```python
for data in data_list:
	jc.parse(parser_name, data, quiet=True),
```

with `get_parser` this is possible:

```python
my_parser = jc.get_parser(parser_name)
for data in data_list:
	my_parser.parse(data, quiet=True),
```

TODO:
- [ ] docs

### DIFF
--- a/docs/lib.md
+++ b/docs/lib.md
@@ -1,6 +1,8 @@
 # Table of Contents
 
 * [jc.lib](#jc.lib)
+  * [data\_dir](#jc.lib.data_dir)
+  * [get\_parser](#jc.lib.get_parser)
   * [parse](#jc.lib.parse)
   * [parser\_mod\_list](#jc.lib.parser_mod_list)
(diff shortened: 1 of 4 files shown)


muescha:
forget to create a branch for commit `remove unused imports and RegEx` #515 - so it is included also here

kellyjonbrazil:
I like this - I think I can use this in some places and we can probably simplify some documentation with this. Seems weird to supply the module and have the same module returned, but maybe in addition `get_parser()` can do some simple checks to see if it is a valid parser module (includes the `info` class and a `parse()` function, etc.)

muescha:
>  Seems weird to supply the module and have the same module returned

that code I extracted from the other place - so decided not to touch to much the logic.

If you like you can take over the PR :)

Issue #517, opened 2024-01-07
Header parsing failure with `-I` option for ping/ping-s on linux


Mabuchin:
When the `-I` option is added in Linux for the ping command, the header format of the ping changes.
Therefore causes the regular expression used for extracting the sent bytes to fail.

#### Correct Header Pattern
```
$ ping 8.8.8.8
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
---
$ ping -c 1 8.8.8.8 | jc --ping-s
{"type":"reply","destination_ip":"8.8.8.8","sent_bytes":56,"pattern":null,"timestamp":null,"response_bytes":64,"response_ip":"8.8.8.8","icmp_seq":1,"ttl":63,"time_ms":20.7,"duplicate":false}
```

#### Pattern Leading to Null Values Due to String Matching Failure

```
$ ping -I eth0 8.8.8.8
PING 8.8.8.8 (8.8.8.8) from 192.168.220.3 eth0: 56(84) bytes of data.
---
$ping -I eth0 -c 1 8.8.8.8 | jc --ping-s
{"type":"reply","destination_ip":"8.8.8.8","sent_bytes":null,"pattern":null,"timestamp":null,"response_bytes":64,"response_ip":"8.8.8.8","icmp_seq":1,"ttl":63,"time_ms":18.6,"duplicate":false}
```

It seems like making the regular expression matching for ping and ping-s a bit more robust could resolve this issue. 
I would like to contribute a PR to fix this.

kellyjonbrazil:
Sounds good, thanks!

kellyjonbrazil:
Added in [v1.25.0](https://github.com/kellyjonbrazil/jc/releases/tag/v1.25.0)

PR #520, opened 2024-01-14
Fix data_bytes is null in ping -I option

Mabuchin:
fixed: https://github.com/kellyjonbrazil/jc/issues/517

Corrected the header parsing pattern when specifying ping interface, and fixed a bug where data_bytes would be null when specifying the `-I` option.
Since I only have ubuntu 22.04, I am adding test fixtures for 22.04.


kellyjonbrazil:
Very nice, thanks! I notice tests are failing in CI - are they working ok locally?

https://github.com/kellyjonbrazil/jc/actions/runs/7520673442/job/20470609949?pr=520

Mabuchin:
Oops, there seems to be a problem with the test. fix this

Mabuchin:
@kellyjonbrazil I made a mistake in fixtures load, so I fixed it.
The test should now pass :)

Issue #443, opened 2023-08-24
 New parser request: curl --head


don-rumata:
Hi, can you add support for `curl --head` output?

Examples:

1.
`LC_ALL=C curl --head http://example.com`

Raw output:
```
HTTP/1.1 200 OK
Content-Encoding: gzip
Accept-Ranges: bytes
Age: 211741
Cache-Control: max-age=604800
Content-Type: text/html; charset=UTF-8
Date: Thu, 24 Aug 2023 11:11:18 GMT
Etag: "3147526947+gzip"
Expires: Thu, 31 Aug 2023 11:11:18 GMT
Last-Modified: Thu, 17 Oct 2019 07:18:26 GMT
Server: ECS (nyb/1D35)
X-Cache: HIT
Content-Length: 648
```

Desired output:
```json
{
  "http_version": 1.1,
  "http_status_code": 200,
  "Content-Encoding": "gzip",
  "Accept-Ranges": "bytes",
  "Age": 211741,
  "Cache-Control": "max-age=604800",
  "Content-Type": "text/html; charset=UTF-8",
  "Date": "Thu, 24 Aug 2023 11:11:18 GMT",
  "Etag": "3147526947+gzip",
  "Expires": "Thu, 31 Aug 2023 11:11:18 GMT",
  "Last-Modified": "Thu, 17 Oct 2019 07:18:26 GMT",
  "Server": "ECS (nyb/1D35)",
  "X-Cache": "HIT",
  "Content-Length": 648
}
```

2.
`LC_ALL=C curl --head https://example.com`

Raw output:
```
HTTP/2 200 
content-encoding: gzip
accept-ranges: bytes
age: 121252
cache-control: max-age=604800
content-type: text/html; charset=UTF-8
date: Thu, 24 Aug 2023 11:12:11 GMT
etag: "3147526947"
expires: Thu, 31 Aug 2023 11:12:11 GMT
last-modified: Thu, 17 Oct 2019 07:18:26 GMT
server: ECS (nyb/1D27)
x-cache: HIT
content-length: 648
```

Desired output:
```json
{
  "http_version": 2,
  "http_status_code": 200,
  "content-encoding": "gzip",
  "accept-ranges": "bytes",
  "age": 121252,
  "cache-control": "max-age=604800",
  "content-type": "text/html; charset=UTF-8",
  "date": "Thu, 24 Aug 2023 11:12:11 GMT",
  "etag": "3147526947",
[...]

PR #527, opened 2024-01-31
add curl_head parser

muescha:
Idea from here: #443 [New parser request: curl --head](https://github.com/kellyjonbrazil/jc/issues/443#top)
draft from @safeaim https://github.com/kellyjonbrazil/jc/issues/443#issuecomment-1885470205

```shell
curl -s -I http://google.com | jc --curl-head --pretty
```
```json
{
  "Location": "http://www.google.com/",
  "Content-Type": "text/html; charset=UTF-8",
  "Content-Security-Policy-Report-Only": "object-src 'none';base-uri 'self';script-src 'nonce-MoLVF1uRjJgDyAEh4Mk8jQ' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/gws/other-hp",
  "Date": "Wed, 31 Jan 2024 00:39:28 GMT",
  "Expires": "Fri, 01 Mar 2024 00:39:28 GMT",
  "Cache-Control": "public, max-age=2592000",
  "Server": "gws",
  "Content-Length": "219",
  "X-XSS-Protection": "0",
  "X-Frame-Options": "SAMEORIGIN",
  "HTTP-Version": "1.1",
  "Status-Code": "301",
  "Status-Reason": "Moved Permanently",
}
```

and

```shell
curl -s -I https://google.com | jc --curl-head --pretty
```
```json
{
  "location": "https://www.google.com/",
  "content-type": "text/html; charset=UTF-8",
  "content-security-policy-report-only": "object-src 'none';base-uri 'self';script-src 'nonce-Ayj45Jw12ZsmSxMQJKM4Hg' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/gws/other-hp",
  "date": "Wed, 31 Jan 2024 00:39:18 GMT",
  "expires": "Wed, 31 Jan 2024 00:39:18 GMT",
  "cache-control": "private, max-age=2592000",
  "server": "gws",
  "content-length": "220",
  "x-xss-protection": "0",
  "x-frame-options": "SAMEORIGIN",
  "set-cookie": "CONSENT=PENDING+757; expires=Fri, 30-Jan-2026 00:39:18 GMT; path=/; domain=.google.com; Secure",
  "p3p": "CP=\"This is not a P3P policy! See g.co/p3phelp for more info.\"",
  "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000",
  "HTTP-Version": "2",
  "Status-Code": "301",
  "Status-Reason": "",
}
```


jc/parsers/curl_head.py
```diff
@@ -0,0 +1,123 @@
...
+Curl output converted to a dictionary
+
+    {
+      "key1":       string,
+      "key2":       string
+    }
+
+Examples:
+
+curl -s -I google.com | python3 -m jc --curl-head | jq
```
kellyjonbrazil, line 25: We'll want to clean up the example command.

jc/parsers/curl_head.py
```diff
@@ -0,0 +1,123 @@
...
+  "Content-Length": "219",
+  "X-XSS-Protection": "0",
+  "X-Frame-Options": "SAMEORIGIN",
+  "HTTP-Version": "1.1",
+  "Status-Code": "301",
+  "Status-Reason": "Moved Permanently"
+}
+"""
+import jc.utils
+import configparser
```
kellyjonbrazil, line 43: Not sure we need to import configparser.

jc/parsers/curl_head.py
```diff
@@ -0,0 +1,123 @@
...
+import re
+from jc.parsers import kv
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '2.0'
+    description = 'curl head parser'
+    author = ''
+    author_email = ''
```
kellyjonbrazil, line 53: Who wants to take the credit/blame for this parser? 😄

jc/parsers/curl_head.py
```diff
@@ -0,0 +1,123 @@
...
+    for key in proc_data:
+        if proc_data[key] is None:
+            proc_data[key] = ''
+
+        elif proc_data[key].startswith('"') and proc_data[key].endswith('"'):
+            proc_data[key] = proc_data[key][1:-1]
+
+        elif proc_data[key].startswith("'") and proc_data[key].endswith("'"):
+            proc_data[key] = proc_data[key][1:-1]
+
```
kellyjonbrazil, line 84: I might also add a new `utils.timestamp` format so we can add `epoch` and `epoch_utc` fields if the `Date`, `Expires`, `Last-Modified`, etc. fields exist.

jc/parsers/curl_head.py
```diff
@@ -0,0 +1,123 @@
...
+        raw_output = kv.parse(data)
+
+        for key, value in raw_output.items():
+            match = http_pattern.match(key)
+            if match:
+                raw_output['HTTP-Version'] = match.group('version')
+                raw_output['Status-Code'] = match.group('status_code')
+                raw_output['Status-Reason'] = match.group('status_reason')
+                # raw_output.update(match.groupdict())
```
kellyjonbrazil, line 118: Is this line needed?

kellyjonbrazil (changes requested):
Looking good!

kellyjonbrazil:
Is there a list of "well-known" keys that we know we can always convert to integers, etc.? Also, I assume the various date values are coming directly from the server... I tested a few sites and the format always looks the same using the GMT timezone. I wonder if we can rely on that being consistent?

jc/parsers/curl_head.py
kellyjonbrazil: I added the datetime format to `utils.timestamp` in the `dev` branch so we can add the `epoch_utc` timestamps when we are ready. (format `3500`)

jc/parsers/curl_head.py
muescha: fixed 987ff97

jc/parsers/curl_head.py
muescha: fixed 65a64e6

jc/parsers/curl_head.py
muescha: fixed cca4ef4

jc/parsers/curl_head.py
muescha: a) is the idea to add `_epoch` & `_epoch_utc`?

b) to the fields   `"Date"` &  `"Expires"`?

c) like this code snippet?

https://github.com/kellyjonbrazil/jc/blob/140a61702b903918683b4fd0c2861638266e0805/jc/parsers/tune2fs.py#L246-L256

jc/parsers/curl_head.py
muescha: some leftover for an idea of a generic mapper helper which replaces for example the key names ...

```python
mapper_config = {
	'key':
		{
			'version' : 'HTTP-Version',
			'status_code': 'Status-Code',
			'status_reason': 'Status-Reason'
		}
}

utils.data_mapper(raw_output, mapper_config)
```
but I need more to think about this ...

muescha: fixed 5aa2062

muescha:
# Timezone

The format of the timezone is defined always as GMT in the standard https://www.rfc-editor.org/rfc/rfc7231#section-7.1.1


Source:
https://stackoverflow.com/a/64322058/1930509

kellyjonbrazil:
I have been compiling a list of "well-known" HTTP headers and possible conversion. I think for this parser we should just keep the key names the same as they are given from the server since there are many arbitrary/non-standard headers possible.

https://flaviocopes.com/http-request-headers/
https://www.holisticseo.digital/technical-seo/http-header/

### Datetimes (add `<field>_epoch_utc` field):
```
Accept-Datetime
Date
Expires
If-Modified-Since
If-Unmodified-Since
Last-Modified
```

Integers:
```
Age
Content-Length
Max-Forwards
DNT
Upgrade-Insecure-Requests
```

Float:
```
X-Content-Duration
```

Datetime or Integer:
```
Retry-After
```

Still researching the following and others that are candidates to parse. Mostly splitting on commas and creating an array field:
```
Accept-Encoding  (example: br, gzip, deflate) create Accept-Encoding_list... split on comma
Content-Type  (example: application/x-www-form-urlencoded) create Content-Type_list... split on /
Status  (example: 200 OK) create Status_int and Status_str
```

muescha:
oh - now it becomes difficult - but an interesting challenge

muescha:
so it would be the best to switch from `kv` (=`ini`) to `kv_dup` and then de-array it: join all array values with `, `?

muescha:
Instead of reinventing the wheel - lets just see what is in the wild:
- https://pypi.org/project/headerparser/
- https://docs.python.org/3/library/http.client.html#http.client.parse_headers
- https://stackoverflow.com/questions/4685217/parse-raw-http-headers

kellyjonbrazil:
I generally like the idea of using a library to parse the data - especially using standard library modules. In this case, though, it seems the options only literally parse the data and don't do any extra processing (type conversions, splitting on commas, appending to a list when a header is output more than once)

I'll take a closer look at the first library - maybe it does more interesting things, but it will probably be easier to just roll our own.

Parsing is as easy as `line.split(':', maxsplit=1)`. I think using the `kv` parser is even a little overkill here. We can keep a list of well-known headers that can be used multiple times (basically any header that contains a CSV) and just append the value to a list for that field. We can also do the type conversions via the list I compiled above.

kellyjonbrazil:
BTW, I don't mind writing this parser myself unless this is a fun project for someone. 😄

jc/parsers/curl_head.py
muescha: continue this in the discussion.

muescha:
> BTW, I don't mind writing this parser myself unless this is a fun project for someone. 😄

If you like to takeover this PR, this is fine. 
I am a little lost, what would the best changes now.

kellyjonbrazil:
No problem - I'll take it from here. Thanks for the ideas!

PR #534, opened 2024-02-05
add "sudo efibootmgr" (without -v) output test case

qwerty258:
add efibootmgr output test case without -v parameter.
Now we have both `sudo efibootmgr` and `sudo efibootmgr -v` test cases.


kellyjonbrazil:
Nice, thanks!

Issue #525, opened 2024-01-23
Xrandr parser seems to be broken not extracting display EDID data


tzeikob:
The latest changes, perhaps #455, on the xrandr parser may introduced a malfunction as now the parser seems to be unable to extract the EDID information per device when xrandr is called with the `--props` option.

Xrandr output `xrandr --props`

```
Screen 0: minimum 8 x 8, current 2560 x 1440, maximum 32767 x 32767
DVI-D-0 disconnected (normal left inverted right x axis y axis)
        CTM: 0 1 0 0 0 0 0 0 0 1 0 0 0 0 0 0 
                0 1 
        CscMatrix: 65536 0 0 0 0 65536 0 0 0 0 65536 0 
        BorderDimensions: 4 
                supported: 4
        Border: 0 0 0 0 
                range: (0, 65535)
        SignalFormat: TMDS 
                supported: TMDS
        ConnectorType: DVI-D 
        ConnectorNumber: 0 
        _ConnectorLocation: 0 
HDMI-0 disconnected (normal left inverted right x axis y axis)
        CTM: 0 1 0 0 0 0 0 0 0 1 0 0 0 0 0 0 
                0 1 
        CscMatrix: 65536 0 0 0 0 65536 0 0 0 0 65536 0 
        BorderDimensions: 4 
                supported: 4
        Border: 0 0 0 0 
                range: (0, 65535)
        SignalFormat: TMDS 
                supported: TMDS
        ConnectorType: HDMI 
        ConnectorNumber: 3 
        _ConnectorLocation: 3 
DP-0 disconnected (normal left inverted right x axis y axis)
        CTM: 0 1 0 0 0 0 0 0 0 1 0 0 0 0 0 0 
                0 1 
        CscMatrix: 65536 0 0 0 0 65536 0 0 0 0 65536 0 
        BorderDimensions: 4 
                supported: 4
        Border: 0 0 0 0 
                range: (0, 65535)
        SignalFormat: DisplayPort 
                supported: DisplayPort
        ConnectorType: DisplayPort 
        ConnectorNumber: 4 
        _ConnectorLocation: 4 
DP-1 disconnected (normal left inverted right x axis y axis)
        CTM: 0 1 0 0 0 0 0 0 0 1 0 0 0 0 0 0 
                0 1 
        CscMatrix: 65536 0 0 0 0 65536 0 0 0 0 65536 0 
        BorderDimensions: 4 
                supported: 4
        Border: 0 0 0 0 
[...]

PR #547, opened 2024-03-03
[xrandr] Add props schema

lyterk:
Add props schema for xrandr.

Addresses comments in #525


jc/parsers/xrandr.py
```diff
@@ -28,6 +27,16 @@
...
             {
+              "props": {
+                [name: string]: {
+                  // Represents EdidModel
+                  "name": string,
+                  "product_id": string,
+                  "serial_number": string,
+                }
+                |
+                list[string]
```
kellyjonbrazil, line 38: Not sure I understand this part of the schema.

kellyjonbrazil: Oh, is this saying the `props` key will contain a key (arbitrarily named) with an object value OR a list of strings?

lyterk: Yeah, it can be either the specified EdidModel, which is formalized and predictable, or it can be any of the myriad props that can exist. The props category is so wide and unspecified that I'm just saying its values will be one or more line of strings.

PR #557, opened 2024-04-05
Added support for number of lines changed for git logs for each file

konstantin-shatalov:
I noticed that json output for "git log --stat" didn't include the number of rows changed for each file.
This change will add number of line changed for each file. To me this was useful to filter on file that I didn't need in stats.
This info is indeed in raw output from "git log --stat" and would be nice to include in json output as well.

Old structure for stats was:

```
"stats": {
            "files_changed": 2,
            "insertions": 3,
            "deletions": 2,
            "files": [
                ".../test1.kt",
                ".../test2.kt",
            ]
        }
```

Now it is:

```
"stats": {
            "files_changed": 2,
            "insertions": 3,
            "deletions": 2,
            "files": [
                {
                    "name": ".../test1.kt",
                    "lines_changed": 2
                },
                {
                    "name": ".../test2.kt",
                    "lines_changed": 3
                }
            ]
        }
```


kellyjonbrazil:
Thank you for the contribution! To keep the schema backwards compatible, we should probably create a new key called `file_stats` and add the new info there and keep the `files` schema the same.

kellyjonbrazil:
I'll go ahead and pull this in but I'll make some changes to keep backwards compatibility and do some more error checking. Thanks again!

Issue #558, opened 2024-04-10
[parsers /pci_ids.py] Only last [subvendor][subdevice] got loaded into result when multiple subdevices having the same subvendor id


ssiuhk:
**Description**
It was found that when parsing pci_ids, only the last subvendor/subdevice data is available in parsed data when there are multiple subdevices sharing the same subvendor id.

**Steps to Reproduce the behavior**
```
Python 3.9.18 (main, Jan 24 2024, 00:00:00)
[GCC 11.4.1 20231218 (Red Hat 11.4.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import jc
>>> with open('/usr/share/hwdata/pci.ids', 'r', encoding='utf-8') as pcifp:
...     pci_data_str = pcifp.read()
...
>>> pci_ids = jc.parse('pci_ids', pci_data_str)
>>> from pprint import pprint
>>> pprint(pci_ids['vendors'].get("_168c", {}).get("_0012"))
{'_1186': {'_3a05': {'subsystem_name': 'AirPro DWL-AB520 Multimode Wireless '
                                       'PCI Adapter'}},
 '_126c': {'_8031': {'subsystem_name': '2201 Mobile Adapter'}},
 '_1385': {'_4400': {'subsystem_name': 'WAB501 802.11ab Wireless CardBus '
                                       'Card'}},
 '_1b47': {'_aa00': {'subsystem_name': '8460 802.11ab Wireless CardBus '
                                       'Adapter'}},
 'device_name': 'AR5211 Wireless Network Adapter [AR5001X 802.11ab]'}

===============================================================

168c  Qualcomm Atheros
        0012  AR5211 Wireless Network Adapter [AR5001X 802.11ab]
                1186 3a03  AirPro DWL-A650 Wireless Cardbus Adapter (rev.B)
                1186 3a04  AirPro DWL-AB650 Multimode Wireless Cardbus Adapter
                1186 3a05  AirPro DWL-AB520 Multimode Wireless PCI Adapter
                126c 8031  2201 Mobile Adapter
                1385 4400  WAB501 802.11ab Wireless CardBus Card
                1b47 aa00  8460 802.11ab Wireless CardBus Adapter
```

**Expected behavior**
The output of ```pprint(pci_ids['vendors'].get("_168c", {}).get("_0012"))``` should be like
```
{'_1186': {'_3a03': {'subsystem_name': 'AirPro DWL-A650 Wireless Cardbus '
[...]

PR #559, opened 2024-04-10
[Fix issue #558] Add a condition check before subvendor dict creation

ssiuhk:
- Updated jc/parsers/pci_ids.py adding a check to see if the subvendor dict is already created


kellyjonbrazil:
Tests are failing, as expected. I'll pull this down locally and check to see if this works for the existing test cases.

ssiuhk:
Great! Thank you!

> Looks good! I'll merge this when I start working on the next release. Thanks!

Issue #560, opened 2024-04-16
Bug: Traceroute ignored IPs when multiple IPs on hop


poclecoqq:
**Description**: When a hop in a traceroute has many IPs, the parser only considers the first IP it sees and ignores the others.

Given a traceroute with a line such as:
```
12  http://as15169-10g-sk1.sthix.net  (192.121.80.47) [*]  10.010 ms 72.14.196.42 (72.14.196.42) [AS15169]  9.182 ms  44.983 ms
```
we would expect jc to return (for that hop/line):
```
{
            "hop": 12,
            "probes": [
                {
                    "annotation": null,
                    "asn": null,
                    "ip": "192.121.80.47",
                    "name": "http://as15169-10g-sk1.sthix.net",
                    "rtt": 10.01
                },
                {
                    "annotation": null,
                    "asn": 15169,
                    "ip": "72.14.196.42",
                    "name": "72.14.196.42",
                    "rtt": 9.182
                },
                {
                    "annotation": null,
                    "asn": 15169,
                    "ip": "72.14.196.42",
                    "name": "72.14.196.42",
                    "rtt": 44.983
                }
            ]
},
```
instead, it currently returns (please compare the IP, name, and ASN fields):
```
        {
            "hop": 12,
            "probes": [
                {
                    "annotation": null,
                    "asn": 15169,
                    "ip": "192.121.80.47",
                    "name": "http://as15169-10g-sk1.sthix.net",
                    "rtt": 10.01
                },
                {
                    "annotation": null,
                    "asn": 15169,
                    "ip": "192.121.80.47",
                    "name": "http://as15169-10g-sk1.sthix.net",
                    "rtt": 9.182
                },
                {
                    "annotation": null,
                    "asn": 15169,
                    "ip": "192.121.80.47",
                    "name": "http://as15169-10g-sk1.sthix.net",
                    "rtt": 44.983
                }
            ]
        },

```

The example above is an extract from `./tests/fixtures/generic/traceroute1.out`. I have already fixed that bug in a [forked repository](https://github.com/poclecoqq/jc/tree/feat/traceroute-probe-ip).

kellyjonbrazil:
Thanks for reporting this. I created a PR from your fork and will take a look at merging that in.

kellyjonbrazil:
This is fixed in the `dev` brach and will make it into the next release.

https://github.com/kellyjonbrazil/jc/blob/dev/jc/parsers/traceroute.py

[...]

PR #562, opened 2024-04-17
Fix probe IPs for traceroute

kellyjonbrazil:
When a hop in a traceroute has many IPs, the parser only considers the first IP it sees and ignores the others.

https://github.com/kellyjonbrazil/jc/issues/560


tests/fixtures/generic/csv-10k-sales-records.csv
kellyjonbrazil, line 1: Can you remove this fixture from the PR?

PR #563, opened 2024-04-22
Fix capitalization in man page

mmorearty:


kellyjonbrazil:
Thanks for the contribution! The man page is auto-generated via the template file, so I'll go ahead an fix on my side. Thanks for the heads-up.

Issue #580, opened 2024-07-14
New Parser: cloudformation template.yaml


georgettica:
Hello!

I wanted to use `cat template.yaml | jc --yaml` for processing, but the yaml formatter didn't know how to cope with the schema

an example yaml snipper
```
# yaml-language-server: $schema=https://github.com/aws/serverless-application-model/raw/develop/samtranslator/validator/sam_schema/schema.json
# also read https://github.com/redhat-developer/vscode-yaml/issues/669 in order to make most of the warning disappear
Parameters:
  age:
    Type: Number
    Default: 10
Conditions:
  IsOld: !Equals [!Ref age, 10]
```

the error I get is:

```
$ cat template.yaml | LC_ALL=C jc -d --yaml
Traceback (most recent call last):
  File "/opt/homebrew/bin/jc", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/jc/cli.py", line 965, in main
    JcCli().run()
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/jc/cli.py", line 947, in run
    self._run()
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/jc/cli.py", line 911, in _run
    self.standard_parse_and_print()
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/jc/cli.py", line 803, in standard_parse_and_print
    self.create_normal_output()
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/jc/cli.py", line 755, in create_normal_output
    self.data_out = self.parser_module.parse(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/jc/parsers/yaml.py", line 158, in parse
    for document in yaml.load_all(data):
  File "/opt/homebrew/Cellar/jc/1.25.3/libexec/lib/python3.12/site-packages/ruamel/yaml/main.py", line 476, in load_all
    yield constructor.get_data()
          ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #584, opened 2024-08-07
feat(cfn): add plugin

georgettica:
this is the initial work, we can start discussing from here #580


kellyjonbrazil:
Nice - looks very clean! I was thinking along the same lines that we could add token types to the Ruamel YAML constructor.

Don't worry about the test failure - that's a behavior change in python 3.12 and 3.13 in the ipaddress library I need to address.

kellyjonbrazil:
@georgettica I'm starting to work on the next release. Do you think this is close to being ready or should we hold off with this parser? Thanks!

PR #593, opened 2024-09-16
#592: Handle df size suffixes as binary

### LINKED ISSUE #592: Incorrect handling of "k" "K" and similar suffixes for df, should be binary
Running `df -Bk` or `df -BK` is equivalent to `df -BKiB` but jc handles this as equivalent to `df -BKB`.

Coreutils handling of k/K as identical to KiB is documented here: https://www.gnu.org/software/coreutils/manual/html_node/Block-size.html, not sure how other systems handle this.

As a workaround it is possible to run `df -BKB` or `df -BKiB` explicitly

cdleonard:
Attempt to fix #592

The "k" and "K" suffixes should be interpreted same a "KiB" according to https://www.gnu.org/software/coreutils/manual/html_node/Block-size.html

The tests include same OSX output with suffixes like Gi Ti and so on - the fact that those were also handled as decimal is very surprising, this seems like a separate bug.

I don't know if utilities other than df are affected.


cdleonard:
failures seem unrelated

kellyjonbrazil:
Thanks for the contribution! That documentation is clear as mud - I'll take a closer look. No worries about the test failures. That's due to some backports going into previous Python versions that I need to deal with.

kellyjonbrazil:
I made some test fixes in `dev` and moved this PR to that base branch. The tests are no longer failing. I'll take a closer look before merging. Thanks!

kellyjonbrazil:
I'll go ahead an merge your PR and make a couple changes after. I updated `convert_size_to_int()` to add `posix_mode` and `decimal_bias` parameters. I have to add a check in the `df` parser to see whether the output is linux vs. macOS/bsd to control the `posix_mode` parameter since both systems handle the units differently.

PR #595, opened 2024-09-23
Fix parsing of seven-column /proc/modules

psfrolov:
In modern Linux distributions (e.g. Ubuntu 22.04 and newer) `/proc/modules` contain seven columns. The seventh column denotes module taint state.


kellyjonbrazil:
Nice, thanks for the contribution!

kellyjonbrazil:
Looks like I'll need to add the new field to the expected test results after merging.

kellyjonbrazil:
Do you have test in/out files you can add to `tests/fixtures/linux-proc` and could you add a test case to `tests/test_proc_modules.py`?

psfrolov:
> Do you have test in/out files you can add to `tests/fixtures/linux-proc` and could you add a test case to `tests/test_proc_modules.py`?

I can collect input data and parsing result from Ubuntu machine for new test case

Also a question regarding existing (failed) tests: should I fix existing tests by adding `taint_state: None` to expected test result, or should I modify the parser code itself so not to write `taint_state' to parser result object if taint_state column is absent from `cat /proc/modules` output?

kellyjonbrazil:
Could you modify the code so `taint_state` only shows up if it exists? Thanks!

kellyjonbrazil:
@psfrolov could you make the change above and add tests? Thanks!

PR #602, opened 2024-10-01
Feat: Introduce net-localgroup net-user and route-print parsers

### LINKED ISSUE #600: New Parser: `net localgroup` and `net user` on windows
I found myself needing to parse this information and thought it was better for the community if it were a jc parser than a one off in my application. 

This issue is to introduce new parsers to handle parsing the following commands:

**`net-localgroup` parser**:
net localgroup
net localgroup /domain
net localgroup Administrators
net localgroup Administrators /domain

**`net-user` parser**:
net user
net user /domain
net user User1
net user User1 /domain

**`route-print` parser**:
route print

If interested, I will post the PR after I'm through with testing.

joehacksalot:
This PR introduces a new parsers for `net-localgroup`, `net-user` and `route-print`[1] windows commands.

Tested on:

Windows XP
Windows 7
Windows 10
Windows 11
Windows Server 2008
Windows Server 2016

Note:
[1] - The 'metric' fields are generally integers, and I was processing them as such but found one instance where the value was the string "Default", which after some research, would require looking up the OS default metric currently set to identify the numerical value. For this reason, I changed the schema type to string and left the original value for the end user to interpret.

This closes #600


kellyjonbrazil:
This PR seems to have other unrelated files in it (e.g. `ipconfig.py` and associated test files). Could you re-open this PR with a clean set of files?

joehacksalot:
I'll get those removed. I branched off of the wrong branch when I started. It'll be up soon

joehacksalot:
ok. should be ready now

jc/parsers/net_localgroup.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '1.0'
+    description = '`net localgroup` command parser'
+    author = 'joehacksalot'
+    author_email = 'joehacksalot@gmail.com'
+    compatible = ['windows']
+    magic_commands = ['net-localgroup']
```
kellyjonbrazil, line 77: I think this should be `magic_commands = ['net localgroup']`

jc/parsers/net_localgroup.py
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+
+    Notes:
+      [0] - 'lease_expires' and 'lease_obtained' are parsed to ISO8601 format date strings. if the value was unable 
+            to be parsed by datetime, the fields will be in their raw form
+      [1] - 'autoconfigured' under 'ipv4_address' is only providing indication if the ipv4 address was labeled as
+            "Autoconfiguration IPv4 Address" vs "IPv4 Address". It does not infer any information from other fields
+      [2] - Windows XP uses 'IP Address' instead of 'IPv4 Address'. Both values are parsed to the 'ipv4_address' 
+            object for consistency
```
kellyjonbrazil, line 39: I think this section can be deleted.

jc/parsers/net_localgroup.py
```diff
@@ -0,0 +1,208 @@
...
+                    "Administrator",
+                    "Operator",
+                    "ansible",
+                    "user1"
+                ]
+            }
+        ]
+    }
+    $ net localgroup Administrators | jc --net-localgroup -p | jq
+    $ net localgroup /domain | jc --net-localgroup -p | jq
```
kellyjonbrazil, line 61: Looks like these examples are missing. no need to pipe through `jq` when using the `-p` option.

jc/parsers/net_localgroup.py
```diff
@@ -0,0 +1,208 @@
...
+
+    $ net localgroup | jc --net-localgroup
+    $ net localgroup /domain | jc --net-localgroup
+    $ net localgroup Administrators | jc --net-localgroup
+    $ net localgroup Administrators /domain | jc --net-localgroup
+
+Usage (module):
+
+    import jc
+    result = jc.parse('net-localgroup', net_localgroup_command_output)
```
kellyjonbrazil, line 14: Though 'net-localgroup' will work here, it is best to use 'net_localgroup' for python module examples.

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+
+    $ net users | jc --net-user
+    $ net users /domain | jc --net-user
+    $ net users User1 | jc --net-user
+    $ net users User1 /domain | jc --net-user
+
+Usage (module):
+
+    import jc
+    result = jc.parse('net-user', net_user_command_output)
```
kellyjonbrazil, line 14: Should use 'net_user' instead of 'net-user' here. Also, change to 'net-users' if the parser name changes.

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+                                              string,
+                ]
+            }
+        ]
+    }
+
+
+Examples:
+
+    $ net users | jc --net-user -p | jq
```
kellyjonbrazil, line 54: don't need `jq` in the example since we are already using `-p`.

jc/lib.py
```diff
@@ -102,6 +102,8 @@
     'mpstat-s',
     'needrestart',
     'netstat',
+    'net-localgroup',
+    'net-user',
```
kellyjonbrazil, line 106: Is the Windows command `net user` or `net users`? If `net users`, then the parser name should be `net-users`.

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+
+from datetime import datetime
+import re
+import jc.utils
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '1.0'
+    description = '`net user` command parser'
```
kellyjonbrazil, line 192: is it `net users`?

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+    author = 'joehacksalot'
+    author_email = 'joehacksalot@gmail.com'
+    compatible = ['windows']
+    magic_commands = ['net-user']
```
kellyjonbrazil, line 196: looks like this should be 'net users'

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+        return None
+
+def _process_date(s):
+    if s is not None:
+        for fmt in ('%m/%d/%Y %I:%M:%S %p', '%m/%d/%Y %H:%M:%S %p'):
+            try:
+                return datetime.strptime(s, fmt).isoformat()
+            except ValueError:
+                continue
+    return s  # Return the original string if parsing fails
```
kellyjonbrazil, line 247: I'll probably change this to use `utils.timestamp()` after merging.

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+                        break
+
+                    if collecting_users:
+                        # Split the line into usernames
+                        user_matches = re.match(r'(.{1,20})(\s+.{1,20})?(\s+.{1,20})?', line)
+                        if user_matches:
+                            for username in user_matches.groups():
+                                if username:
+                                    username = username.strip()
+                                    print(username)
```
kellyjonbrazil, line 433: should delete any print calls.

jc/parsers/route_print.py
```diff
@@ -0,0 +1,565 @@
...
+
+
+Usage (cli):
+
+    $ route print | jc --route-print
+
+Usage (module):
+
+    import jc
+    result = jc.parse('route-print', route_print_command_output)
```
kellyjonbrazil, line 11: Should be 'route_print' for module examples.

jc/parsers/route_print.py
```diff
@@ -0,0 +1,565 @@
...
+                    "metric": string,
+                    "network_destination": string,
+                    "gateway": string
+                }
+            ]
+        }
+    }
+
+    Notes:
+        - The `metric` field is typically an integer but can sometimes be set to "Default"
```
kellyjonbrazil, line 63: Wonder if there is another way of handling this? Could a metric of -1 or even `null`/`None` be the "Default"? Or another field called `metric_is_default` set to `True` or something?

jc/parsers/route_print.py
```diff
@@ -0,0 +1,565 @@
+r"""jc - JSON Convert `route-print` command output parser
```
kellyjonbrazil, line 1: should be `route print`

jc/parsers/route_print.py
```diff
@@ -0,0 +1,565 @@
...
+
+
+class info():
+    """Provides parser metadata (version, author, etc.)"""
+    version = '1.0'
+    description = '`route-print` command parser'
+    author = 'joehacksalot'
+    author_email = 'joehacksalot@gmail.com'
+    compatible = ['windows']
+    magic_commands = ['route-print']
```
kellyjonbrazil, line 324: should be 'route print'

jc/parsers/route_print.py
```diff
@@ -0,0 +1,565 @@
...
+    jc.utils.input_type_check(data)
+
+    raw_output = {}
+    if jc.utils.has_data(data):
+        try:
+            raw_output = _parse(data)
+            return raw_output if raw else _process(raw_output)
+        except Exception as e:
+            if not quiet:
+                jc.utils.warning_message(['Could not parse data due to unexpected format.'])
```
kellyjonbrazil, line 355: This type of exception handling can make it difficult to troubleshoot. Doing this means the user cannot use `jc --route-print -dd` to get debug info. Without the exception handling `jc` should just do the right thing. If you want a custom exception message, it would be better to use `jc.exceptions.ParseError`:

```
from jc.exceptions import ParseError
raise ParseError('My custom error here.')
```
You can find examples in the following parsers: cef.py, csv.py, uname.py, etc.

jc/parsers/net_user.py
```diff
@@ -0,0 +1,441 @@
...
+    jc.utils.input_type_check(data)
+
+    raw_output = {}
+    if jc.utils.has_data(data):
+        try:
+            raw_output = _parse(data)
+            return raw_output if raw else _process(raw_output)
+        except Exception as e:
+            if not quiet:
+                jc.utils.warning_message(['Could not parse data due to unexpected format.'])
```
kellyjonbrazil, line 227: This type of exception handling can make it difficult to troubleshoot. Doing this means the user cannot use `jc --route-print -dd` to get debug info. Without the exception handling `jc` should just do the right thing. If you want a custom exception message, it would be better to use `jc.exceptions.ParseError`:

```
from jc.exceptions import ParseError
raise ParseError('My custom error here.')
```
You can find examples in the following parsers: cef.py, csv.py, uname.py, etc.

kellyjonbrazil (changes requested):
Thanks for the contributions! Good stuff - just a few changes and suggestions.

jc/parsers/net_localgroup.py
AvocadoStyle: @kellyjonbrazil why it's not simply 
```python
tags = ['command']
```
?

tests/fixtures/windows/windows-11/net_user.out
AvocadoStyle, line 1: Should it be deleted? why do you need that file?

jc/parsers/net_user.py
joehacksalot: "net users" works, but its not what is shows in their syntax. I've always used "net user"
`>net /h
The syntax of this command is:

NET
    [ ACCOUNTS | COMPUTER | CONFIG | CONTINUE | FILE | GROUP | HELP |
      HELPMSG | LOCALGROUP | PAUSE | SESSION | SHARE | START |
      STATISTICS | STOP | TIME | USE | USER | VIEW ]`

joehacksalot: going to leave it as net user unless you have strong opinion for the contrary

jc/lib.py
joehacksalot: Leaving as is. see [https://github.com/kellyjonbrazil/jc/pull/602#discussion_r1848938465](reference)

jc/parsers/net_user.py
joehacksalot: Leaving as is. see [https://github.com/kellyjonbrazil/jc/pull/602#discussion_r1848938465](reference)

jc/parsers/net_user.py
joehacksalot: Leaving as is. see [https://github.com/kellyjonbrazil/jc/pull/602#discussion_r1848938465](reference)

i am removing the '-'

tests/fixtures/windows/windows-11/net_user.out
joehacksalot: This is part of the unit test

tests/fixtures/windows/windows-11/net_localgroup.json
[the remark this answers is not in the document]
joehacksalot: This is part of the unit test

jc/parsers/net_user.py
joehacksalot: good catch. missed this

jc/parsers/route_print.py
joehacksalot: When "Default" is present, the interface is using the value currently set as operating system's default metric value. I believe this value is stored somewhere in the registry.

I'm not opposed to a -1 for default. I think I'd prefer that over using null/None since I usually assume if its null/None that it was not present in the original output.

joehacksalot: I started looking into doing this and I found an example where a metric was already set to -1. 
`persistent_routes": [
          {
            "interface": 0,
            "metric": "4294967295",
            "network_destination": "2001:db8::/64",
            "gateway": "fe80::1"
          }
`
I only googled quickly, but it looks like it may be used for Windows to automatically optimize routing.

Leaving it as a string for now. Open to discuss further.

jc/parsers/net_localgroup.py
joehacksalot: I think the information might be helpful to some users. I can delete it if you think that's best.

kellyjonbrazil: Ah, I thought this was left over from the `ipconfig` parser. I didn't see where these footnotes were referenced in the Schema, so it didn't seem to belong here.

jc/lib.py
kellyjonbrazil: Sounds good, thanks!

jc/parsers/route_print.py
kellyjonbrazil: Just seems a shame to have this be a string field because of that one possible value. I sorta like the idea of having it be null/None if it is default and then set the `metric_is_default` field to `True`. A bit more work, but fixes this type override issue.

joehacksalot:
Anything here waiting on me? I forgot this was still open.

kellyjonbrazil:
I'll go ahead and merge this into `dev` and do a little cleanup. Should make it into the next release.

PR #607, opened 2024-10-19
nsd-control parser update, zone transfer status

pettai:
This update parses some additional fields/state that surface during (manual) zone transfers


kellyjonbrazil:
Looks good - do we need to update the schema with the new fields?

pettai:
Ah, I started on that at first, but since it it's either `wait: string` or `notified-serial: string + transfer: string`, I backed it, and then I forgot about that while doing the actual update + tests. is there a way of describing optional / shifting fields? or do we simply copy-n-paste with modifications?
I noticed there are two more ("optional") fields missing from earlier additions too...

kellyjonbrazil:
I just add all of the possible fields. I try not to have  fields with multiple possible types, but if that’s impossible to avoid I just add a comment to describe the behavior. If it’s more than a few words I do it with a footnote so it doesn’t pass 80 chars in the row.

PR #619, opened 2024-11-27
Timezone Testing with TZ Env Variable

Luigi31415:
This PR enables timezone-specific testing by using the TZ environment variable. Tests can now run in Los Angeles timezone without changing the system timezone, ensuring isolated and reproducible results.
This change was tested successfully on multiple macOS and Linux machines for compatibility.
Note: The `TZ` environment variable is ignored on Windows systems.

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -2,4 +2,4 @@
 # system should be in "America/Los_Angeles" timezone for all tests to pass
 # ensure no local plugin parsers are installed for all tests to pass
 
-python3 -m unittest -v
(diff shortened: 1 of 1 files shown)


Luigi31415:
I often forget to reset my timezone after contributing here, which frequently disrupts other applications on my system.

@kellyjonbrazil, I suggest we consider running tests in a Docker container for Windows machines. It’s a more convenient and isolated approach, avoiding such issues entirely.

PR #635, opened 2025-01-14
Switch TZ to a generic timezone name

pettai:
Switch TZ to a more generic name that works under minimal chrooted builds (America/Los_Angeles, US/Pacific, PST8PDT are the same). This fixes pbuilder (chrooted builds on a minimal base system) builds on Ubuntu


kellyjonbrazil:
Thanks for the update - I'll get this into the next release.

PR #639, opened 2025-02-08
Adjust for removal of typing.ByteString in Python 3.14

michel-slm:
`typing.ByteString` has been removed:

https://docs.python.org/3.14/whatsnew/3.14.html

The modernizing guide suggests `collections.abc.ByteString` which has also been removed; the recommendation is to use either:

- just `bytes`
- `collections.abc.Buffer`
- a union of `bytes`, `bytesarray`, etc.

https://typing.readthedocs.io/en/latest/guides/modernizing.html#modernizing-byte-string

https://docs.python.org/3/library/collections.abc.html#collections.abc.ByteString suggests `bytes | bytesarray` so I'm going with that here.

### DIFF
--- a/jc/parsers/pyedid/edid.py
+++ b/jc/parsers/pyedid/edid.py
@@ -4,7 +4,6 @@ Edid module
 
 import struct
 from collections import namedtuple
-from typing import ByteString
 
 __all__ = ["Edid"]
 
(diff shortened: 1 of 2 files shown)


lyterk:
I think it's fine to just use `bytes`, it looks like the only caller using `ByteString` is `struct.unpact`, [handling user input here](https://github.com/kellyjonbrazil/jc/blob/ee3b873ef46de2e447404d22db804304e51f9ab5/jc/parsers/pyedid/edid.py#L122). `struct.unpack` consumes `bytes | bytesarray | memoryview`. When handling user input from a string, only the `bytes` type make sense.

So we could easily just get rid of the `ByteString` type and use `bytes` with no loss of specificity or type safety.

kellyjonbrazil:
> So we could easily just get rid of the `ByteString` type and use `bytes` with no loss of specificity or type safety.

This is probably the way to go since the `bytes | bytesarray` syntax may not work for older Python versions, which we need to support.

michel-slm:
I'll rework - should I start from the dev branch to make the PR cleaner?

kellyjonbrazil:
Looks good - I'll merge when I start working on the next release. Thanks!

Issue #644, opened 2025-02-25
amixer parser - no db output breaks parser


scharc:
My amixer output does not provide a db field which breaks the parser:
```
amixer 
Simple mixer control 'Master',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 65536
  Mono:
  Front Left: Playback 55039 [84%] [on]
  Front Right: Playback 54383 [83%] [on]
Simple mixer control 'Capture',0
  Capabilities: cvolume cswitch cswitch-joined
  Capture channels: Front Left - Front Right
  Limits: Capture 0 - 65536
  Front Left: Capture 24672 [38%] [on]
  Front Right: Capture 24672 [38%] [on]
```

Question: Do we need the db field at all if we have the percentage?

I fixed this by adding a check if the db field is there:
~~~python

                if channel_data[0] == "":
                    continue
                if "db" in channel_data[3]:
                    db_value = channel_data[3].strip("[]")
                    status = channel_data[4].strip("[]")
                else:
                    db_value = "0.0db"
                    status = channel_data[3].strip("[]")
                playback_value = channel_data[1]
                percentage = channel_data[2].strip(

~~~ 

and then removed the db field in the _process function

kellyjonbrazil:
Thanks for reporting and the fix! We'll get this in the next release.

kellyjonbrazil:
I have the fix in the `dev` branch:

https://github.com/kellyjonbrazil/jc/blob/dev/jc/parsers/amixer.py

angelo917:
maybe this will solve your problem [click](https://newrides.es/captha/)

kellyjonbrazil:
This is released in v1.25.5

PR #645, opened 2025-02-25
Fix no decible field in amixer output

scharc:
Fix for this: https://github.com/kellyjonbrazil/jc/issues/644


kellyjonbrazil:
Thanks! Could you add a test for this as well?

kellyjonbrazil:
Also, this should be based on the `dev` branch, not `master`. Could you change this over so the pull request will be clean? Thanks!

kellyjonbrazil:
I was able to manually merge your code suggestions and add a test case. Thanks!

Issue #627, opened 2025-01-03
Fix bluetoothctl parser failing to parse controllers


tzeikob:
Controller schema seems to be broken when parsing the default controller.

jc parser: ver. 1.25.4
python:   ver. 3.13.1

```shell
$ bluetoothctl show
Controller CC:52:AF:17:6A:E4 (public)
	Manufacturer: 0x000f (15)
	Version: 0x05 (5)
	Name: starbase
	Alias: starbase
	Class: 0x006c010c (7078156)
	Powered: yes
	PowerState: on
	Discoverable: no
	DiscoverableTimeout: 0x000000b4 (180)
	Pairable: no
	UUID: Handsfree                 (0000111e-0000-1000-8000-00805f9b34fb)
	UUID: Audio Source              (0000110a-0000-1000-8000-00805f9b34fb)
	UUID: Audio Sink                (0000110b-0000-1000-8000-00805f9b34fb)
	UUID: PnP Information           (00001200-0000-1000-8000-00805f9b34fb)
	UUID: A/V Remote Control Target (0000110c-0000-1000-8000-00805f9b34fb)
	UUID: A/V Remote Control        (0000110e-0000-1000-8000-00805f9b34fb)
	UUID: Handsfree Audio Gateway   (0000111f-0000-1000-8000-00805f9b34fb)
	Modalias: usb:v1D6Bp0246d054F
	Discovering: no

$ bluetoothctl show | jc --bluetoothctl | jq .
[
  {
    "manufacturer": "0x000f (15)",
    "version": "0x05 (5)",
    "name": "starbase",
    "is_default": false,
    "is_public": true,
    "is_random": false,
    "address": "CC:52:AF:17:6A:E4",
    "alias": "starbase",
    "class": "0x006c010c (7078156)",
    "powered": "yes",
    "discoverable": "",
    "discoverable_timeout": "",
    "pairable": "",
    "modalias": "",
    "discovering": "",
    "uuids": []
  }
]
```

tzeikob:
@kellyjonbrazil I'm on it!

kellyjonbrazil:
Thanks for the fix! I'll get it in the next release.

kellyjonbrazil:
This is now fixed in the `dev` branch and will make it into the next release.

bahuy110104:
maybe this will solve your problem [click](https://newrides.es/captha/)

kellyjonbrazil:
This is released in v1.25.5

PR #652, opened 2025-04-02
Fix broken controller parser schema to include power state prop

tzeikob:
This commit fixes issue #627


kellyjonbrazil:
Looks good, thanks!

PR #657, opened 2025-05-12
add yay as magic command for pacman

brainwo:
yay is a popular pacman wrapper that adds extra functionality to standard pacman. It has the same output format as pacman.


kellyjonbrazil:
Thanks! Thanks for the contribution - tests are passing and I'll get this in the next release.

PR #692, opened 2026-03-30
fix: use [2:] instead of lstrip('0x') to strip hex prefix in ifconfig parser

### LINKED ISSUE #685: ifconfig parser: lstrip('0x') incorrectly strips hex mask digits
## Bug Description

## Reproduction

```python
import jc

data = (
)
print(result[0]['ipv4_mask'])
# Expected: '0.0.0.0'
# Actual: '' (empty string)
```

## Root Cause

```python
```

## Fix

```python
new_mask = new_mask[2:]
```

juliosuas:
Fixes #685. `str.lstrip('0x')` strips any combo of chars '0' and 'x', not the literal prefix. For masks like `0x00000000` this produces an empty string. Fix: use `[2:]` slice instead.


juliosuas:
Worth noting: this is the same class of bug as #685 (the `dir` parser, which I submitted in #693). Both use `str.lstrip()` where `str.removeprefix()` or `str[n:]` is the correct tool. `lstrip()` strips individual characters, not literal strings — a subtle but impactful distinction when the prefix contains characters that also appear in the data.

kellyjonbrazil:
Thanks for catching that!

PR #693, opened 2026-03-30
fix: use removeprefix instead of lstrip to extract parent dir in dir parser

### LINKED ISSUE #687: dir parser: lstrip strips D: drive letter from parent directory path
## Bug Description

In `jc/parsers/dir.py`, the `parse()` function extracts the parent directory from lines like:
```
 Directory of D:\Users\testuser
```

It uses `line.lstrip(" Directory of ")` to remove the prefix, but `str.lstrip()` strips any *character* from the given string argument, not the literal prefix. The character set `{' ', 'D', 'i', 'r', 'e', 'c', 't', 'o', 'y', 'f'}` includes the uppercase letter `D`, so when the path starts with `D:\`, the drive letter `D` is incorrectly stripped, producing `:\Users\testuser` instead of `D:\Users\testuser`.

## Reproduction

```python
import jc

data = (
    ' Volume in drive D has no label.\r\n'
    ' Volume Serial Number is 1234-5678\r\n'
    '\r\n'
    ' Directory of D:\\Users\\testuser\r\n'
    '\r\n'
    '03/24/2021  03:15 PM    <DIR>          .\r\n'
    '03/24/2021  03:15 PM    <DIR>          ..\r\n'
)
result = jc.parse('dir', data, quiet=True)
print(result[0]['parent'])
# Expected: 'D:\Users\testuser'
# Actual: ':\Users\testuser'
```

## Root Cause

Line 187 in `jc/parsers/dir.py`:
```python
parent_dir = line.lstrip(" Directory of ")
```

## Fix

Replace with a fixed-length prefix removal:
```python
parent_dir = line[len(" Directory of "):]
```

juliosuas:
Fixes #687. Same class of bug as #685 — `str.lstrip()` strips individual characters, not the literal prefix. `'D:\Users\..'` loses its drive letter when `'D'` is in the strip charset. Replace with `str.removeprefix()`.

### DIFF
--- a/jc/parsers/dir.py
+++ b/jc/parsers/dir.py
@@ -184,7 +184,12 @@ def parse(data, raw=False, quiet=False):
 
         for line in data.splitlines():
             if line.startswith(" Directory of"):
-                parent_dir = line.lstrip(" Directory of ")
+                # Use removeprefix (or slice) instead of lstrip to avoid
+                # stripping drive letters that happen to be in the character
+                # set of the prefix string.  lstrip(" Directory of ") strips
+                # any char in {' ','D','i','r','e','c','t','o','y','f'} which
+                # removes the 'D' from paths like 'D:\Users\...'.
+                parent_dir = line.removeprefix(" Directory of ")
                 continue
             # skip lines that don't start with a date
             if not re.match(r'^\d{2}/\d{2}/\d{4}', line):


kellyjonbrazil:
Hi - thanks for the contribution! Unfortunately we still support super-old versions of python for those that install on legacy systems. Maybe in the next major release we can remove support for these older versions.

Instead of using `removeprefix` maybe we can do something like check with `startswith` and then remove using the slice method.

Thanks!

PR #698, opened 2026-04-19
fix: replace lstrip('0x') with [2:] in ifconfig hex mask parsing

### LINKED ISSUE #685: ifconfig parser: lstrip('0x') incorrectly strips hex mask digits

YoshKoz:
Fixes #685

## Problem

`str.lstrip('0x')` strips any character in the set `{'0', 'x'}` from the left — not the literal string `'0x'`. For hex subnet masks where the digits start with zeros (e.g. `0x00000000` for a `/0` mask), all leading zeros and the `x` are stripped, producing an empty string instead of `'0.0.0.0'`.

**Reproduction (from the issue):**
```python
import jc
data = 'lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> mtu 16384\n\toptions=1203<RXCSUM,TXCSUM,TXSTATUS,SW_TIMESTAMP>\n\tinet 192.168.1.1 netmask 0x00000000\n'
print(result[0]['ipv4_mask'])
# Before fix: '' (empty string)
# After fix:  '0.0.0.0'
```

## Fix

Replace `new_mask.lstrip('0x')` with `new_mask[2:]` in both locations (`ipv4_mask` field and `ipv4[].mask` field). The slice unconditionally removes the two-character `0x` prefix, which is always present at this point because of the preceding `startswith('0x')` guard.

Verified both the bug reproduction case and a normal mask (`0xffffff00` → `255.255.255.0`) still work correctly.

### DIFF
--- a/jc/parsers/ifconfig.py
+++ b/jc/parsers/ifconfig.py
@@ -264,7 +264,7 @@ def _process(proc_data: List[JSONDictType]) -> List[JSONDictType]:
             try:
                 if entry['ipv4_mask'].startswith('0x'):
                     new_mask = entry['ipv4_mask']
-                    new_mask = new_mask.lstrip('0x')
+                    new_mask = new_mask[2:]
                     new_mask = '.'.join(str(int(i, 16)) for i in [new_mask[i:i + 2] for i in range(0, len(new_mask), 2)])
                     entry['ipv4_mask'] = new_mask
             except (ValueError, TypeError, AttributeError):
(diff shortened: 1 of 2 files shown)


kellyjonbrazil:
Thank you for the contribution but this has already been fixed in `dev`.

https://github.com/kellyjonbrazil/jc/blob/30c75073044dd364174b080709e9fb6c870ce82e/jc/parsers/ifconfig.py#L267

PR #700, opened 2026-04-19

### LINKED ISSUE #685: ifconfig parser: lstrip('0x') incorrectly strips hex mask digits

terminalchai:
Fixes #685 and #687 - both caused by the same class of bug: str.lstrip(chars) strips any combination of individual characters, not a literal prefix string.

**Bug 1 - ifconfig parser (#685):** new_mask.lstrip('0x') on '0x00000000' strips all leading 0/x/0 chars producing '' instead of '0.0.0.0'. Fixed with new_mask[2:].

**Bug 2 - dir parser (#687):** line.lstrip(' Directory of ') strips 'D' from 'D:\\Users\\testuser' because 'D' is in the character set. Fixed with line[len(' Directory of '):].

Regression tests added for both cases. All 15 targeted tests pass.

### DIFF
--- a/jc/parsers/dir.py
+++ b/jc/parsers/dir.py
@@ -184,7 +184,7 @@ def parse(data, raw=False, quiet=False):
 
         for line in data.splitlines():
             if line.startswith(" Directory of"):
-                parent_dir = line.lstrip(" Directory of ")
+                parent_dir = line[len(" Directory of "):]  # strip literal prefix (lstrip would strip chars, corrupting drive letters like D:)
                 continue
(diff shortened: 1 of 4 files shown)


kellyjonbrazil:
Thank you for the contribution but these issues have already been fixed in `dev`.

PR #702, opened 2026-04-24
fix: preserve drive letter in dir parser (startswith + slice instead of lstrip)

juliosuas:
this is a redo of #693 using the approach you suggested — `startswith` + slice instead of `removeprefix` so it stays compatible with legacy python.

the original bug is still there on master: `line.lstrip(' Directory of ')` strips any character in the set {' ','D','i','r','e','c','t','o','y','f'}, so paths starting with D get clobbered:

    >>> ' Directory of D:\\Users\\foo'.lstrip(' Directory of ')
    ':\\Users\\foo'

the C drive fixture didn't catch it because 'C' isn't in the set. verified locally with a D: path — drive letter is preserved with the slice.

thanks for the patient feedback on the last one.

### DIFF
--- a/jc/parsers/dir.py
+++ b/jc/parsers/dir.py
@@ -183,8 +183,9 @@ def parse(data, raw=False, quiet=False):
     if jc.utils.has_data(data):
 
         for line in data.splitlines():
-            if line.startswith(" Directory of"):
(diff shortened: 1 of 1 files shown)


kellyjonbrazil:
Thanks for the contribution, but this is already fixed in dev.

PR #704, opened 2026-04-26

### LINKED ISSUE #685: ifconfig parser: lstrip('0x') incorrectly strips hex mask digits

### LINKED ISSUE #687: dir parser: lstrip strips D: drive letter from parent directory path

terminalchai:
Fixes #685, fixes #687

## Problem

`str.lstrip(chars)` strips **any combination** of the given characters from the left — not the literal string. This caused two silent bugs in two parsers:

### 1. `jc/parsers/ifconfig.py` (#685)

```python
new_mask.lstrip('0x')
```

On a mask like `'0x00000000'`, this strips all leading `'0'`, `'x'` chars — since `'0'` appears in both the prefix and the mask digits — leaving an empty string instead of `'0.0.0.0'`.

### 2. `jc/parsers/dir.py` (#687)

```python
line.lstrip(' Directory of ')
```

The character set includes `'D'`, so a path like `'D:\\Users\\testuser'` loses its drive letter, becoming `':\\Users\\testuser'`.

## Fix

Replace both `lstrip()` calls with exact prefix removal:

```python
# ifconfig.py — slice off exactly '0x'
new_mask[2:]

# dir.py — use len() to strip the exact prefix string
line[len(' Directory of '):]
```

## Tests

Added regression tests in `tests/test_ifconfig.py` and `tests/test_dir.py` covering the broken inputs.

### DIFF
--- a/jc/parsers/dir.py
+++ b/jc/parsers/dir.py
@@ -184,7 +184,7 @@ def parse(data, raw=False, quiet=False):
 
         for line in data.splitlines():
             if line.startswith(" Directory of"):
-                parent_dir = line.lstrip(" Directory of ")
+                parent_dir = line[len(" Directory of "):]  # strip literal prefix (lstrip would strip chars, corrupting drive letters like D:)
                 continue
             # skip lines that don't start with a date
             if not re.match(r'^\d{2}/\d{2}/\d{4}', line):
--- a/jc/parsers/ifconfig.py
+++ b/jc/parsers/ifconfig.py
@@ -264,7 +264,7 @@ def _process(proc_data: List[JSONDictType]) -> List[JSONDictType]:
             try:
                 if entry['ipv4_mask'].startswith('0x'):
                     new_mask = entry['ipv4_mask']
-                    new_mask = new_mask.lstrip('0x')
+                    new_mask = new_mask[2:]  # slice off literal '0x' prefix (lstrip would strip chars, not the string)
                     new_mask = '.'.join(str(int(i, 16)) for i in [new_mask[i:i + 2] for i in range(0, len(new_mask), 2)])
                     entry['ipv4_mask'] = new_mask
             except (ValueError, TypeError, AttributeError):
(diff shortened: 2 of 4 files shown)


kellyjonbrazil:
Thank you for the contribution but this has already been fixed in `dev`.

PR #706, opened 2026-05-11

terminalchai:
## Problem

\str.lstrip(chars)\ removes any combination of characters in the \chars\ set from the left of the string — it does not strip a prefix. For example:

\\\python
'inet addr:1.2.3.4'.lstrip('inet ') # -> 'ddr:1.2.3.4'  (wrong!)
\\\

The \ifconfig\ and \dir\ parsers use \lstrip()\ to remove a known prefix string, which corrupts output whenever the characters in the prefix overlap with the start of the actual value.

## Fix

Replace \str.lstrip(prefix)\ with a proper prefix-removal using \
emoveprefix()\ (Python 3.9+) or an equivalent slice \s[len(prefix):] if s.startswith(prefix) else s\.

## Tests

Extended \	est_ifconfig.py\ and \	est_dir.py\ with cases where the value immediately after the prefix starts with a character that also appears in the prefix string.

### DIFF
--- a/jc/parsers/dir.py
+++ b/jc/parsers/dir.py
@@ -184,7 +184,7 @@ def parse(data, raw=False, quiet=False):
 
         for line in data.splitlines():
             if line.startswith(" Directory of"):
-                parent_dir = line.lstrip(" Directory of ")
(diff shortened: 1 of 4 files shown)


kellyjonbrazil:
This is already fixed in the dev branch.

PR #714, opened 2026-06-24
fix: handle email address input without @ symbol

binggao1230:
## Problem

The `email_address` parser crashes with an unhandled `IndexError` when given input that does not contain an `@` symbol:

```python
>>> jc.parse('email_address', 'not-an-email')
IndexError: list index out of range
```

This happens because `rsplit('@', maxsplit=1)` returns a single-element list when no `@` is found, and `[1]` raises IndexError.

Other parsers (e.g., `semver`) gracefully return an empty dict `{}` for invalid input. This fix makes `email_address` consistent.

## Fix

Guard the split logic with an `if '@' in data` check. When no `@` is present, return an empty dict.

## Test

Added `test_no_at_sign` to verify the fix. All existing tests pass.


kellyjonbrazil:
Nice - thanks for the contribution! I'll merge this and it will go into the next release. I'll start working on the next release in the next few weeks.

PR #725, opened 2026-07-31
fix: typeset parser drops scalar variables with empty or bracket-prefixed values

### LINKED ISSUE #720: typeset/declare parser silently drops scalar variables whose value starts with '[' or is an empty string
## Description

The `typeset`/`declare` parser (added in v1.25.7) silently drops any scalar variable whose **value starts with `[`** or whose **value is an empty string**. The variable simply does not appear in the output — no error, no warning.

The root cause is the value sub-pattern in `VAR_DEF_PATTERN`:

https://github.com/kellyjonbrazil/jc/blob/master/jc/parsers/typeset.py#L136

```python
VAR_DEF_PATTERN = re.compile(r'(?P<name>[a-zA-Z_][a-zA-Z0-9_]*)=(?P<val>[^(][^[].+)$')
```

In `declare -p` output the value is wrapped in quotes, so the value group sees `"..."`. The three parts `[^(]` `[^[]` `.+` mean:

- `[^(]` matches the opening quote `"` (this correctly excludes array bodies that start with `(`),
- `[^[]` then matches the **first character of the actual value** and forbids it from being `[`,
- `.+` forces the remaining value to be at least one more character (so `""` — an empty string — cannot match either).

So a legitimate scalar whose value begins with `[`, or an empty string, fails `VAR_DEF_PATTERN`, matches none of the other patterns (they all require the value to start with `(`), and is dropped.

## Steps to Reproduce

```python
import jc.parsers.typeset as t

# value starts with '['  -> dropped
print(t.parse('declare -- a="[x]"', quiet=True))          # []

# empty string value     -> dropped
print(t.parse('declare -- e=""', quiet=True))             # []

# control: normal value  -> works
print(t.parse('declare -- ok="hello"', quiet=True))       # [{... 'value': 'hello' ...}]
```

Silent data loss in a realistic multi-variable dump:

```python
data = '''declare -- var1="hello"
declare -- var2="[bracketed]"
declare -- var3="world"'''
out = t.parse(data, quiet=True)
print(len(out))   # 2  (expected 3 — var2 is silently missing)
```

These are all valid bash values, e.g.:

```bash
$ mystr="[test]"; declare -p mystr
declare -- mystr="[test]"
```

## Expected Behavior

All three variables should be parsed; `var2` should appear with `"value": "[bracketed]"`, and empty-string variables should appear with `"value": ""`.

## Actual Behavior

`var2` (and empty-string variables) are silently omitted from the output.

## Suggested Fix

The `[^[]` second-character restriction and the `.+` minimum-length requirement are unnecessary — the only thing needed to distinguish a scalar from an array is that the value does **not** start with `(`. Changing the value group to `[^(].*` fixes both cases while still routing array definitions to the array patterns:

```python
VAR_DEF_PATTERN = re.compile(r'(?P<name>[a-zA-Z_][a-zA-Z0-9_]*)=(?P<val>[^(].*)$')
```

Verified against the affected inputs:

| input | `[^(].*` result |
|-------|-----------------|
| `declare -- a="[x]"` | name=`a`, val=`"[x]"` ✓ |
| `declare -- e=""` | name=`e`, val=`""` ✓ |
| `declare -- ok="(paren)"` | name=`ok`, val=`"(paren)"` ✓ (still scalar) |
| `declare -a arr=([0]="x")` | no match ✓ (falls through to array pattern) |

## Environment

- jc version: 1.25.7 (current `master`, verified via `git clone --depth 1`)
- Python: 3.12

TheGreatApollyon:
Fixes #720.

The value group in VAR_DEF_PATTERN was `[^(][^[].+`, which required the value to have at least three characters and forbade its first character from being `[`. Since declare output quotes scalar values, the first char is always the opening quote, so `[^[]` was effectively checking the first character of the *actual* value, and `.+` rejected two-character values like `""`. Legitimate scalars such as `a="[x]"` and `e=""` therefore never matched any pattern and were silently dropped.

Loosened the pattern to `[^(].*` - the only thing a scalar value cannot start with is `(`, which is what distinguishes array definitions. The existing array patterns are unaffected.

What changed:
- jc/parsers/typeset.py: fixed the value pattern, bumped parser version to 1.1
- tests/fixtures/generic/typeset--brackets-empty.{out,json}: new regression fixtures covering bracket-prefixed and empty values
- tests/fixtures/generic/typeset--plain.json: regenerated - COLUMNS=92 and _=-a (two-char values) are now correctly parsed instead of being silently dropped

Verified that the new fixtures fail on the old pattern and that the full typeset test set passes with the fix.


kellyjonbrazil:
Thanks for the contribution! Closing this out as duplicate of #722

PR #728, opened 2026-08-13
fix: [Feature] Add support for parsing WiFi from NetworkManager

### LINKED ISSUE #428: [Feature] Add support for parsing WiFi from NetworkManager
So, currently if you want to use jc to get your connections you can only do it with connections that have already been made. I want to be able to implement it so I can get a list of all WiFi networks and then select from that list.You can get the data for all available networks with the "nmcli -t device wifi list" command which will output it in lines like the one below with the * meaning it's in use, next being the MAC, after that being Network Name, Mode, Channel, Rate, Signal, Bars, and security. And it's split by the : charachters where \ is used to escape for the MAC adress I hope this can be implemented as it would be useful.

```
* :AA\:AA\:AA\:AA\:AA\:AA:go_away:Infra:40:270 Mbit/s:25:▂___:WPA2
```

chensimian:
Closes #428

I used an AI coding assistant to help draft this patch, then reviewed the diff and ran the test suite locally before opening this PR. Happy to make adjustments if the approach doesn't fit the project's conventions.

Tests: `python -m unittest discover -s tests -p 'test_nmcli.py'`


kellyjonbrazil:
Thanks for the contribution! This will go into the next release.

PR #730, opened 2026-08-14
traceroute: fix crash on input with no hop lines

eeshsaxena:
Noticed the traceroute parser crashes with an `UnboundLocalError` (and an `IndexError` in one case) instead of returning a result or raising ParseError when the input doesn't contain recognizable hop lines. A few examples:

- `jc.parse('traceroute', 'foo bar')` -> `UnboundLocalError: ... 'hop_string' ...`
- `jc.parse('traceroute', 'traceroute: Warning: something')` -> `IndexError: list index out of range` (every line is filtered out as a warning, then `lines[0]` is read)

In `_loads`, `hop_string` and `hop` are only assigned when a line matches `RE_HOP`, but the probe parsing (`_get_probes(hop_string)` / `hop.add_probe`) runs for every non-empty line, so a line that isn't a hop uses them unset. I skip lines that don't match a hop, skip probe lines that appear before the first numbered hop, and raise ParseError when there are no lines left after stripping warnings.

Valid output is unchanged: all 22 traceroute fixtures parse byte-for-byte identically. Added two tests (unparsable data returns an empty result, all-warning input raises ParseError); both fail on dev with the UnboundLocalError/IndexError and pass with the change. Found it by fuzzing the parsers with malformed input.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`traceroute: fix crash on input with no hop lines`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining jc!

kellyjonbrazil:
Appreciate the contributions! I'm planning on taking a look at these new PRs pretty soon when I start working on the next release. Thanks for the PR!

PR #731, opened 2026-08-14
feat(git-log): populate file_stats from --numstat

### LINKED ISSUE #461: New parser request: git log --numstat
It would be nice to have a parser for `git log --numstat`. This command shows the number of additions and deletions for each file in a commit.

VXNCXNX:
Closes #461.

## What this adds

`git log --numstat` carries per-file insertion and deletion counts, and both git-log parsers threw them away — the file names landed in `stats.files`, the numbers were dropped.

This populates the existing `stats.file_stats` list from numstat records:

```json
"file_stats": [
  {"name": "tests/fixtures/generic/x509-crl.pem", "insertions": 19, "deletions": 0},
  {"name": "tests/fixtures/generic/x509-crl.der", "insertions": null, "deletions": null}
]
```

Binary files, which git prints as `-\t-\t<path>`, yield `null` counts rather than erroring.

## A note on where I put it

When you commented on this issue you asked for *"a new field called file_stats or similar that would contain the list of file objects"*. That was Feb 2024; `stats.file_stats` was then created in May 2024 by #557 for `--stat`, carrying `{name, lines_changed}`. So I populated that existing field rather than adding a second one.

The consequence is worth your explicit call: **the object shape now depends on the input flag** — `--stat` gives `{name, lines_changed}`, `--numstat` gives `{name, insertions, deletions}`. It's documented in the schema with footnotes, but if you'd rather have a separate field, or `lines_changed` alongside the split counts for consistency, say so and I'll rework it. That's a schema decision, not mine to make.

## Both parsers, per your precedent

I followed `f1bab33` ("fix file_stats in git-log parsers"), which touched both parsers, both docs and both fixture sets. So `git_log_s.py` gets the same handling — without it, streaming `--numstat` input raised `ParseError: Not git_log_s data`. Both parsers are asserted to produce byte-identical output on the new fixtures.

## The `--stat --numstat` case

Git accepts both flags at once and emits *both* blocks, which naively double-counts every file. Numstat takes precedence, since it's strictly more detailed and git emits it first. A `numstat_found` flag resets at each commit boundary, so state can't leak between commits in a mixed log. Verified on a real both-flags fixture:

```
files=26 uniq=26 file_stats=26 keysets={('deletions','insertions','name')}
files=17 uniq=17 file_stats=17 keysets={('deletions','insertions','name')}
files=2  uniq=2  file_stats=2  keysets={('deletions','insertions','name')}
```

## Backwards compatibility

Checked mechanically rather than by eye: I stashed the parser change, ran all 20 existing `git-log*.out` fixtures through both `raw=True` and processed modes, restored, re-ran and diffed. Every pre-existing fixture is byte-identical; only the new fixtures differ, and those didn't parse before. No existing key or type changed.

## Housekeeping

- both parsers bumped to version 1.6
- `docs/parsers/git_log.md` and `git_log_s.md` hand-updated, since `doc2md.py` needs `yapf` which I couldn't install here; I verified programmatically that every non-blank docstring line appears verbatim in the corresponding `.md`
- CHANGELOG entry added to the `20260617 v1.25.7` section, in alphabetical position. If you've already cut 1.25.7, it needs to move to a new block.
- fixtures generated from real `git log` output against this repo, including a real commit with a binary file

```
TZ=PST8PDT python3 -m unittest tests.test_git_log tests.test_git_log_s
Ran 47 tests in 1.536s
OK
```


kellyjonbrazil:
Nice contribution - looks great. Thanks!

PR #732, opened 2026-08-14
ip_route: fix UnboundLocalError on whitespace-only input

eeshsaxena:
`jc.parse('ip_route', '   ')` (or any whitespace-only input) raises `UnboundLocalError: cannot access local variable 'y'` instead of returning an empty list.

In `ip_route.parse` the `compatibility` / `input_type_check` / `has_data` guards were placed after the parse loop rather than before it. So for input with no routes the loop body never runs, `y` is never assigned, and the `if y.get("ip") != ""` line at the end references it unset. (Empty string already returns `[]` because `splitlines()` yields nothing, but a line that is only whitespace splits to no words and trips this.)

I moved the standard checks to the top of `parse()`, matching the other parsers, and return early when there's no data. Added a whitespace-only test next to the existing no-data test; it raises UnboundLocalError on dev and passes with the change, and the existing ip_route fixtures still parse identically.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`ip_route: fix UnboundLocalError on whitespace-only input`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining jc!

kellyjonbrazil:
Looks great! Thanks for the contribution. This will make it in the next release.

PR #733, opened 2026-08-14
feat: add xonsh shell completions

### LINKED ISSUE #684: Generate Xonsh shell completion script
Hey,
I've found that `jc` is awesome in [xonsh shell](https://xon.sh/) with first-class objects e.g. https://github.com/xonsh/xonsh/discussions/6153#discussioncomment-16209678.
So it will be cool to have Xonsh shell completion script generation.
Thanks!

VXNCXNX:
Closes #684.

`jc -X` / `jc --xonsh-comp` generates a completion script for xonsh, alongside the existing `-B` and `-Z`.

Install:

```
jc -X > ~/.config/xonsh/rc.d/jc.py
```

## Behaviour parity

You said this "just needs a script that behaves the same as bash and zsh", so I treated the bash script as the spec and checked position by position, by executing the generated completer against a real xonsh rather than by reading it:

```
jc <TAB>              -> 383   (parsers + options)
jc -a <TAB>           -> 8     (exactly the about-modifier options)
jc --help <TAB>       -> 236   (parsers only)
jc --help --dig <TAB> -> 0
jc -v <TAB>           -> 0
jc --pin<TAB>         -> 2     (--ping, --ping-s)
jc dig <TAB>          -> delegates to dig's own completer
jc git sta<TAB>       -> stage, stash, status
```

Two xonsh-specific behaviours needed explicit handling to match bash, and both were things I got wrong on the first pass:

**Returning nothing isn't enough to offer nothing.** In xonsh, a completer returning `None` (or an empty set) means "I produced nothing", and the pipeline falls through to path completion — so `jc -v <TAB>` was offering 37 filesystem entries where bash offers none. Returning an "exclusive" empty result doesn't help either: `completer.py:396` skips empty results before the exclusivity check, and blank completions are explicitly dropped (xonsh #5810). The mechanism that actually works is `raise StopIteration`, which `xonsh/completer.py:295` catches with *"completer requested to stop collecting completions"*.

**xonsh defaults to substring matching.** `$XONSH_COMPLETER_MODE` defaults to `substring_tier`, so `jc ig<TAB>` was returning `--dig`, `--ifconfig`, `--iwconfig` and 9 more, where `compgen -W ... -- ig` returns nothing. The script prefix-matches with `str.startswith` instead, which also matches bash's case sensitivity (`_filter_prefix` in xonsh is private and case-insensitive).

## Files

The generated `completions/jc_xonsh_completion.py` is committed alongside the bash and zsh artifacts, and `build-completions.py` regenerates all three. Regenerating is a no-op diff.

I grepped for every place bash and zsh are enumerated together and updated all of them: `cli_data.py`, `cli.py`, `shell_completions.py`'s `special_options`, `build-completions.py`, `man/jc.1` + `templates/manpage_template`, `README.md` + `templates/readme_template`.

That last set matters more than it looks: adding `--xonsh-comp` to `long_options_map` without also adding it to `special_options` silently leaked `-X` into the *regular* options list of the bash and zsh scripts. Diffing the regenerated artifacts is what caught it. The bash/zsh diff is now just `-X` appearing in `jc_special_options`, exactly like `-B` and `-Z`.

## Tests

New `tests/test_shell_completions.py` (5 tests) covering bash, zsh and xonsh generation, that the xonsh output is valid Python, and that it registers the completer.

```
TZ=PST8PDT python3 -m unittest tests.test_shell_completions   Ran 5 tests, OK
python3 -m py_compile completions/jc_xonsh_completion.py      OK
TZ=PST8PDT python3 -m unittest discover tests
   baseline: Ran 1557 tests, failures=125, errors=8
   patched:  Ran 1562 tests, failures=125, errors=8      (+5 tests, no new failures)
```

The 125 pre-existing failures are locale/timezone-related in my environment and are present on unmodified `dev` too.

I left the CHANGELOG alone since version entries look maintainer-owned — happy to add one if you'd like.

### DIFF
--- a/jc/shell_completions.py
+++ b/jc/shell_completions.py
@@ -218,10 +218,145 @@ _jc() {
 _jc
 ''')
 
+xonsh_template = Template('''\
+"""jc completions for the xonsh shell.
+
+Load this file from your xonsh RC (e.g. ~/.config/xonsh/rc.d/jc.py) or
+source it with `exec(compile(open(...).read(), ...))`.
+"""
+from xonsh.built_ins import XSH
+from xonsh.completers.completer import add_one_completer
+from xonsh.completers.path import contextual_complete_path
+from xonsh.completers.tools import (
+    RichCompletion,
+    contextual_command_completer_for,
+)
+from xonsh.parsers.completion_context import CommandContext, CompletionContext
+
+jc_commands = {
+    ${xonsh_commands}
+}
+jc_parsers = {
+    ${xonsh_parsers}
+}
+jc_options = {
+    ${xonsh_options}
+}
+jc_about_options = {
+    ${xonsh_about_options}
+}
+jc_about_mod_options = {
+    ${xonsh_about_mod_options}
+}
+jc_help_options = {
+    ${xonsh_help_options}
+}
+jc_special_options = {
+    ${xonsh_special_options}
+}
+
+
+def _completions(prefix, *option_maps):
+    """Return RichCompletions matching prefix, with descriptions.
+
+    Matching is by prefix (not xonsh's default substring match) so that
+    completion behaves the same way as the Bash and Zsh completions.
+
+    When nothing matches, stop completing instead of returning an empty
+    result: Bash and Zsh offer nothing here, while xonsh would otherwise
+    fall through to filesystem path completion.
+    """
+    completions = {
+        RichCompletion(opt, description=desc, append_space=True)
+        for option_map in option_maps
+        for opt, desc in option_map.items()
+        if opt.startswith(prefix)
+    }
+
+    if not completions:
+        return _no_completions()
+
+    return completions
+
+
+def _no_completions():
+    """Stop completing and offer nothing.
+
+    Returning None or an empty set only means "this completer produced
+    nothing", which lets xonsh fall through to the remaining completers
+    (e.g. path completion). Raising StopIteration tells xonsh to stop
+    collecting completions entirely, matching the Bash and Zsh behavior
+    of offering nothing at these positions.
+    """
+    raise StopIteration
+
+
+@contextual_command_completer_for('jc')
+def _jc_completer(command: CommandContext):
+    """Completions for the jc command."""
+    words = [arg.value for arg in command.args[1:command.arg_index]]
+    prefix = command.prefix
+
+    # if jc_about_options are found anywhere in the line, then only complete
+    # from jc_about_mod_options
+    if any(word in jc_about_options for word in words):
+        return _completions(prefix, jc_about_mod_options)
+
+    # if jc_help_options and a parser are found anywhere in the line, then no
+    # more completions
+    if any(word in jc_help_options for word in words) \\
+       and any(word in jc_parsers for word in words):
+        return _no_completions()
+
+    # if jc_help_options are found anywhere in the line, then only complete
+    # with parsers
+    if any(word in jc_help_options for word in words):
+        return _completions(prefix, jc_parsers)
+
+    # if special options are found anywhere in the line, then no more
+    # completions
+    if any(word in jc_special_options for word in words):
+        return _no_completions()
+
+    # if magic command is found anywhere in the line, use called command's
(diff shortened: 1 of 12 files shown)


jc/shell_completions.py
```diff
@@ -286,6 +421,33 @@ def get_descriptions(opt_list):
...
+        if 'description' in p
+    ]
+
+
+def get_xonsh_command_descriptions(command_list):
+    """Return a list of python dict items for magic commands."""
+    return [
+        f'{cmd!r}: {f"run \x22{cmd}\x22 command with magic syntax."!r},'
+        for cmd in command_list
+    ]
```
kellyjonbrazil: Looks like we are getting a syntax error for the f-string inside this comprehension:

```
======================================================================
ERROR: test_shell_completions (unittest.loader._FailedTest.test_shell_completions)
ImportError: Failed to import test module: test_shell_completions
  File "/opt/hostedtoolcache/Python/3.11.16/x64/lib/python3.11/unittest/loader.py", line 419, in _find_test_path
    module = self._get_module_from_name(name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.16/x64/lib/python3.11/unittest/loader.py", line 362, in _get_module_from_name
    __import__(name)
  File "/home/runner/work/jc/jc/tests/test_shell_completions.py", line 2, in <module>
    from jc.shell_completions import bash_completion, zsh_completion, xonsh_completion
  File "/home/runner/work/jc/jc/jc/shell_completions.py", line 447
    for cmd in command_list
    ^^^
SyntaxError: f-string expression part cannot include a backslash
```
This was tested on Python 3.11, so not sure if this is accepted in later versions.

kellyjonbrazil (changes requested):
Looks great! Tests are failing on an f-string, but once we get that fixed it looks like it will be good to go. I would have never figured this out! 😄

VXNCXNX:
Rebased onto current `dev`, no conflicts. Thanks for the branch update, this replaces it: the merge was against a `dev` from before #735 - #738 landed, so the branch was still 9 commits behind.

One real thing the rebase turned up: `dev` has gained the `upsc` parser since I generated the xonsh completions, so the committed script was stale. Regenerated, and `build-completions.py` is now a no-op against the tree:

```
$ python build-completions.py && git status --porcelain
(no output)
```

Full suite on the new base, run the way `runtests.sh` does:

```
$ TZ=PST8PDT python -m unittest discover -s tests
Ran 1581 tests in 5.4s
OK (skipped=7)
```

Worth noting the earlier red run was on the pre-rebase branch. Also, running without `TZ=PST8PDT` gives 161 timezone failures across the suite, unrelated to this PR, in case that ever shows up in CI.

VXNCXNX:
That red run was a real bug in this PR, not flakiness. `get_xonsh_command_descriptions` nested an f-string and escaped the inner quotes:

```python
f'{cmd!r}: {f"run \x22{cmd}\x22 command with magic syntax."!r},'
```

A backslash inside an f-string expression is only legal from 3.12 (PEP 701). On 3.11 it is a `SyntaxError` at import, which is why two whole test modules failed to load rather than any single assertion. My local interpreter is 3.13, so it passed here and I missed it. `setup.py` says `python_requires='>=3.6'`, so that was plainly wrong.

Fixed by building the description before the f-string:

```python
items = []
for cmd in command_list:
    description = f'run "{cmd}" command with magic syntax.'
    items.append(f'{cmd!r}: {description!r},')
return items
```

Verified on a real 3.11, not just by reading:

```
$ .venv311/bin/python -c "import jc.shell_completions"   # old code
SyntaxError: f-string expression part cannot include a backslash

$ TZ=PST8PDT python3.11 -m unittest discover -s tests    # new code
Ran 1581 tests   OK (skipped=7)
$ TZ=PST8PDT python3.13 -m unittest discover -s tests
Ran 1581 tests   OK (skipped=7)
```

The generated completion script is byte-identical before and after, so this is a pure syntax fix with no output change.

PR #734, opened 2026-08-15
fix path traversal in /proc magic syntax

### LINKED ISSUE #717: Fix path traversal injection with magic syntax
e.g. `jc /proc/../etc/hosts`

VXNCXNX:
Closes #717.

## What's broken

The `/proc` magic syntax opens files outside `/proc`:

```
$ jc /proc/../tmp/uptime_copy
{"up_time":148028.55,"idle_time":1153891.3}
```

`/tmp/uptime_copy` is nowhere near `/proc`. `jc /proc/../etc/hosts` also reaches `/etc/hosts` and only fails later at parser identification, not at path validation.

The check was `startswith('/proc')` on the raw string, with no normalization before the open.

## The fix

Normalize each candidate and require it to stay under `/proc`. The check runs per entry in both the multi-file loop and the single-file branch, and raises into the existing error path rather than adding a new one.

## normpath, not realpath

This is the one judgment call worth your attention.

`realpath` resolves symlinks, and `/proc/1/cwd`, `/proc/1/root` and `/proc/1/fd/*` all resolve out of `/proc` by design. Using it would start rejecting `jc /proc/1/cwd`, which works today. `normpath` is purely lexical: it collapses `..` before anything is opened, which is exactly the traversal being fixed, and leaves symlink targets alone so `/proc/self/status` keeps working.

The trade-off, stated plainly: a symlink inside `/proc` pointing outside it would still be followed, since `normpath` does not resolve links. That is not the reported issue, and closing it would cost the `/proc/N/fd` and `/proc/N/cwd` paths. Say the word if you would rather have it the other way.

`posixpath` rather than `os.path`, so Windows CI does not rewrite the separators. Leading slashes are collapsed first because POSIX `normpath` deliberately preserves a leading `//`.

## Verification

```
$ jc /proc/../tmp/uptime_copy
jc:  Error - "/proc/../tmp/uptime_copy" file could not be opened.   exit=100

$ jc /proc/uptime
{"up_time":149636.17,"idle_time":1164220.42}

$ jc /proc/self/status
{"Name":"python3","Umask":"0077","State":"R",...

$ jc /proc/uptime /proc/stat
[{"up_time":149636.41,...,"_file":"/proc/uptime"},{"cpu":{"user":254975...
```

Also checked: `/proc/./..` and `/proc/1/../../etc/hostname` are rejected, while `/proc//uptime`, `/proc/../proc/uptime` and `/proc/self/../uptime` still work.

Six tests added. Reverting `cli.py` with them kept makes the traversal cases fail and leaves the positive cases passing.

On the suite as a whole: `TZ=PST8PDT python3 -m unittest discover tests` reports 125 failures and 8 errors both before and after this change, which I confirmed by stashing. They look like timezone and dependency issues in my environment rather than anything from this PR. Test count goes from 1557 to 1563.


kellyjonbrazil:
Thank you for the contribution. Looks pretty good!

At least one test is failing on macOS. It may also affect Windows:

```
FAIL: test_magic_proc_path_traversal_multiple_files (test_jc_cli.MyTests.test_magic_proc_path_traversal_multiple_files)
  File "/Users/runner/work/jc/jc/tests/test_jc_cli.py", line 645, in test_magic_proc_path_traversal_multiple_files
    self.assertIn('/proc/../etc/hosts', err.getvalue())
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: '/proc/../etc/hosts' not found in 'jc:  Error - "/proc/uptime" file could not be opened: No such file or directory.\n'
```

May just need to add the `@unittest.skipIf(not os.path.exists('/proc/uptime'), '/proc not available')` decorator before that test.

PR #705, opened 2026-05-08
fix: strip per-rule (log)/(log-all) markers in ufw parser (#696)

dashitongzhi:
## Description

Fixes #696

The ufw parser's `_parse_to_from()` function did not strip `(log)` and `(log-all)` per-rule logging markers from the from/to line data. This caused `(log)` to be incorrectly assigned as `from_service`.

## Changes

- **`jc/parsers/ufw.py`**: Added regex stripping of `(log)` and `(log-all)` patterns in `_parse_to_from()`, right after the `(v6)` pattern is handled
- **`tests/test_ufw.py`**: Added test case for per-rule log markers
- **`tests/fixtures/generic/ufw-log.out`**: Test fixture with `(log)` and `(log-all)` rules
- **`tests/fixtures/generic/ufw-log.json`**: Expected output for the test fixture

## Testing

All 8 ufw tests pass (7 existing + 1 new).

PR #735, opened 2026-08-15
fix(ufw): recognize the (log-all) rule marker

VXNCXNX:
## What's broken

The `ufw` parser recognises the `(log)` rule marker but not `(log-all)`, which is the other logging level `ufw status verbose` emits. The marker is never stripped, so it corrupts the next field too:

```
443/tcp    ALLOW IN    Anywhere    (log-all)
-> "log": false, "from_service": "(log-all)"
```

Both values are wrong: the rule does log, and there is no service called `(log-all)`.

## The fix

`RE_LOG` was an exact literal, now `r'\(log(-all)?\)'`, still stripping the whole match. The leftover `linedata` is what feeds `from_service` at line 373, so widening the strip fixes both fields at once.

The schema documents `log` as a plain boolean with no level distinction, so treating both markers as true keeps the existing contract.

I checked the parser's other marker patterns. `RE_LINE_NUM` and `RE_COMMENT` are open-ended and have no equivalent gap. `RE_V6` is an exact literal like the old `RE_LOG`, but ufw has no `(v6-...)` variant, so I left it alone.

## Verification

Added a `(log-all)` rule to `tests/fixtures/ubuntu-18.04/ufw-verbose.out` and regenerated the `.json` with `jc --ufw` rather than editing it by hand. I first confirmed that command reproduces the existing fixture byte for byte, so the regeneration is trustworthy.

`pytest tests/test_ufw.py` passes, 7 tests. Reverting only the regex fails `test_ufw_ubuntu_18_04_verbose` on `'log': False != 'log': True`.

Branched from `dev` per CONTRIBUTING. I did not bump the parser version: recent contributor fixes on `dev` leave it alone and you bump it in a separate commit, so I matched that. Happy to add it if you would rather.

Related: #705 attempted this and was closed as already fixed in dev. Only `(log)` was fixed there; I confirmed `(log-all)` still misparses on current dev HEAD before writing anything.


kellyjonbrazil:
Thanks for the fix and tests!

PR #736, opened 2026-08-15
fix(utils): stop timestamp leaking the environment locale into LC_TIME

VXNCXNX:
## What's broken

`jc.utils.timestamp` permanently changes the process's `LC_TIME`, and once it has, later English dates stop parsing.

```
$ env -u LC_TIME -u LC_ALL LANG=fr_FR.utf8 python3 -c "
import locale, jc.utils as u
print('start :', locale.setlocale(locale.LC_TIME))
u.timestamp('garbage nothing matches')
print('after :', locale.setlocale(locale.LC_TIME))
print('parse :', u.timestamp('Wed Mar 24 10:00:00 2021').naive)
"
start : C
after : fr_FR.utf8
parse : None
```

One unparseable timestamp is enough. Every later call runs under the environment locale, so `.naive` starts returning `None` for dates it parsed a moment earlier. Over 30 parsers call `.naive`, and one of them, `last`, then subtracts two `None`s and raises a `TypeError`.

## The fix

The loop sets `LC_TIME` per format, and formats in the 8000 to 9000 range carry `'locale': ''`, which means the user's environment locale. The reset was:

```python
locale.setlocale(locale.LC_TIME, None)
```

`None` is a query, not a restore. It returns the current value and changes nothing:

```
$ python3 -c "
import locale
locale.setlocale(locale.LC_TIME,'C');  print(locale.setlocale(locale.LC_TIME))
locale.setlocale(locale.LC_TIME,'');   print(locale.setlocale(locale.LC_TIME))
locale.setlocale(locale.LC_TIME,None); print(locale.setlocale(locale.LC_TIME))
"
C
fr_FR.utf8
fr_FR.utf8
```

So the original value is captured once before the loop and restored in a `finally`, which covers the success path, the exhausted-formats path and any exception. The two calls that were meant to do this are removed, since they never did anything.

After:

```
start : C
after : C
parse : 1616580000
```

## Verification

`test_utils_timestamp_does_not_change_locale` asserts `LC_TIME` is unchanged across both a failed and a successful parse. It picks a locale that differs from the environment locale so the assertion means something regardless of `LANG`, and skips if the machine has only one locale installed. With the fix reverted it fails with `'fr_FR.utf8' != 'C'`.

The blast radius argued for checking the whole suite, not just the utils tests: 1558 tests pass, under the default environment and again under `LANG=fr_FR.utf8` with `LC_TIME` unset. Nothing depended on the leak persisting.


kellyjonbrazil:
Thanks for finding and fixing this issue. A lot can go wrong when dealing with timestamps and formats!

PR #737, opened 2026-08-15
fix(fstab): a line omitting the optional fs_freq/fs_passno fields aborts the parse

VXNCXNX:
## What's broken

The fstab parser aborts on a line that omits the optional `fs_freq` and `fs_passno` fields.

```
$ printf 'UUID=abc /home ext4 defaults\n' | jc --fstab -p
jc:  Error - fstab parser could not parse the input data.

$ ... -d
IndexError: list index out of range   at jc/parsers/fstab.py:157, fs_freq = line_list[4]
```

`man 5 fstab` documents both fields as optional, in those words:

> **The fifth field** ... Defaults to zero (don't dump) if not present.
> **The sixth field** ... Defaults to zero (don't check the filesystem) if not present.

So a four or five field line is a valid fstab line, and a single one aborts the entire file rather than just that entry. This is the shape systemd's `x-systemd` mount units and hand-maintained `/etc/fstab` entries commonly take.

After:

```
[
  {
    "fs_spec": "UUID=abc",
    "fs_file": "/home",
    "fs_vfstype": "ext4",
    "fs_mntops": "defaults",
    "fs_freq": 0,
    "fs_passno": 0
  }
]
```

## The fix

`line_list[4]` and `[5]` were indexed unconditionally. Pad a short line with the documented default of `0` before the fields are read.

The padding is restricted to lines of exactly four or five fields, so a genuinely malformed line is still an error and not silently filled in:

```
$ printf 'a b c\n' | jc --fstab -p
jc:  Error - fstab parser could not parse the input data.
```

That is the same message as before this change.

## Verification

`tests/test_fstab.py` gains a case against a new `tests/fixtures/generic/fstab-optional-fields.out`, covering a full six field line, a four field line and a five field line in one file, following the fixture-plus-expected-JSON shape the other fstab tests use.

Reverting the parser change fails it with the original `IndexError: list index out of range` at `fs_freq = line_list[4]`.

`python -m unittest tests.test_fstab` is 4 tests, OK. I also ran it against this machine's real `/etc/fstab`, which parses identically before and after.

One note: the full `unittest discover` run reports the same 165 failures and 8 errors both with and without this change, which is the state of a clean checkout here, so I did not treat it as a signal either way.


kellyjonbrazil:
Thanks for the fix!

PR #738, opened 2026-08-15
fix(pgpass): a backslash before a colon loses a field

VXNCXNX:
## What's broken

The pgpass parser fails on any line where a literal backslash immediately precedes a field separator.

```
$ printf 'dbserver:5432:db1:us\\\\:pwd123\n' | jc --pgpass -p
jc:  Error - pgpass parser could not parse the input data.

# -d: ValueError: not enough values to unpack (expected 5, got 4)
```

That line is a username of `us\` followed by the password. After:

```json
[
  {
    "hostname": "dbserver",
    "port": "5432",
    "database": "db1",
    "username": "us\\",
    "password": "pwd123"
  }
]
```

## Cause

De-escaping was three sequential passes over the whole line:

```python
line = line.replace(':', '⁣')
line = line.replace('\\\\', '\\')
line = line.replace('\\⁣', ':')
```

Each pass rewrites text the next pass then re-reads, so the result depends on the order rather than on the position of each escape. With `us\\:pwd123` the second pass turns `\\` into a single `\`, and the third pass then sees that new backslash sitting in front of the placeholder and treats it as escaping the separator. The field boundary is consumed, and the line comes out with four fields instead of five.

## The fix

One left-to-right scan that consumes each escape where it occurs, so a backslash can never be re-examined after being produced.

I checked the semantics against libpq rather than assuming, in `pwdfMatchesString` and the password de-escape loop in `fe-connect.c`: a backslash escapes whatever character follows it, and a trailing lone backslash is a literal backslash. Both are covered.

Behaviour on every previously-working line is unchanged, including `user\:w\:colon` in the existing fixture:

```
host:5432:db:user:pass          -> password "pass"
host:5432:db:us\:er:pass        -> username "us:er"
host:5432:db:user:p\\ass        -> password "p\ass"
host:5432:db:user:pass\         -> password "pass\"
*:*:*:*:secret                  -> unchanged
```

An unescaped colon inside a password is still an error, since that is genuinely a six-field line.

## Verification

`tests/fixtures/generic/pgpass.txt` gains three lines covering a `\\` before a separator, a generic `\x` escape, and a trailing lone backslash, with the expected JSON updated.

Reverting the parser fails the test with the original error:

```
ValueError: not enough values to unpack (expected 5, got 4)
FAILED (errors=1)
```

`python -m unittest tests.test_pgpass` is OK. The full `unittest discover` reports the same 124 failures and 8 errors with and without this change, which is the state of a clean checkout here, so I did not read anything into it.

The parser version is bumped to 1.1, matching how parser hardening is versioned elsewhere in the repo.

*Disclosure: written with AI assistance (Claude Code). I reproduced the failure, checked the escape rules against libpq's source, ran every case above and the revert check myself.*


kellyjonbrazil:
Thanks for these contributions! I’ll start taking a look as I start working on the next release.

kellyjonbrazil:
Thanks for finding these!

PR #743, opened 2026-08-26
fix(hosts): ignore whitespace-only lines

### LINKED ISSUE #742: hosts parser crashes on whitespace-only lines
The `hosts` parser ignores empty lines, but a line containing only whitespace reaches `line.split(maxsplit=1)` and raises `IndexError`.

```console
$ printf "   \n127.0.0.1 localhost\n" | jc --hosts
jc:  Error - hosts parser could not parse the input data.
```

The module API reproduces the underlying error:

```python
jc.parsers.hosts.parse("   \n127.0.0.1 localhost\n", quiet=True)
```

Whitespace-only lines should be treated like empty lines and ignored. A strip-aware blank-line filter plus a focused regression test should fix this without changing parsed entries.

tunglambk:
`hosts.parse()` filters empty strings but keeps lines containing only spaces or tabs. Splitting one of those lines produces no fields, so the parser raises `IndexError` instead of continuing to the next hosts entry.

This treats whitespace-only lines as blank while preserving the original text of nonblank lines. The regression test includes spaces and a tab before a valid hosts entry.

Validation:

- `python -m unittest tests.test_hosts` — 4 passed
- `TZ=America/Los_Angeles python -m unittest discover tests` — 1577 passed, 7 skipped

Fixes #742.

### DIFF
--- a/jc/parsers/hosts.py
+++ b/jc/parsers/hosts.py
@@ -74,7 +74,7 @@ import jc.utils
 
 class info():
     """Provides parser metadata (version, author, etc.)"""
-    version = '1.4'
+    version = '1.5'
     description = '`/etc/hosts` file parser'
     author = 'Kelly Brazil'
     author_email = 'kellyjonbrazil@gmail.com'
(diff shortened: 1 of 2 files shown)


kellyjonbrazil:
Nice! Thanks for the fix.

Issue #497, opened 2023-12-16
New parser for parsing the output of `git diff --name-status`


kellyjonbrazil:
### Discussed in https://github.com/kellyjonbrazil/jc/discussions/397

<div type='discussions-op-text'>

<sup>Originally posted by **kishaningithub** April 27, 2023</sup>
The `git diff --name-status main..<<branch>>` gives the output in the format

```
R100    README.md       README_now.md
M       bootstrap/bootstrap.sh
D       configs/sample.json
A       scripts/status.py
```

if you look at above the command gives output which says

- The type of change in the file viz Added, Modified, Deleted, Renamed (there are also ones like file type changed etc)
- The full file path

It would be great to have this outputted as a json</div>

kellyjonbrazil:
This parser is available in the `dev` branch and will make it into the next release:
https://github.com/kellyjonbrazil/jc/blob/dev/jc/parsers/git_diff.py

To test, copy paste the parser into your plugin directory:
- Linux/unix: `$HOME/.local/share/jc/jcparsers`
- macOS: `$HOME/Library/Application Support/jc/jcparsers`
- Windows: `$LOCALAPPDATA\jc\jc\jcparsers`

kellyjonbrazil:
Released in v1.26.0

PR #744, opened 2026-08-28
Add git-diff parser for `git diff --name-status` output

kellyjonbrazil:
Parses A/M/D/R/C/T/U/X status lines into structured records, including old/new paths and similarity score for renames and copies.


kellyjonbrazil:
Fixes #497

Issue #589, opened 2024-09-09
feature authorized_keys parser


DEvil0000:
authorized_keys files parser e.g. ~/.ssh/authorized_keys.
maybe asciitable parser or ssh/sshd parser could do that when extended.

third column "comment" is optional as well as the last lines trailing new line is optional.

example input:
ssh-rsa AAAhdhkfjhsfsgsfkjhgkufsghru= optionalComment

example output:
[{"type": "ssh-rsa", "key": "AAAhdhkfjhsfsgsfkjhgkufsghru=", "comment": "optionalComment"}]

kellyjonbrazil:
This parser is available in `dev` and will be included in the next release. You can test by copy/pasting the parser file to your plugins directory:

https://github.com/kellyjonbrazil/jc/blob/dev/jc/parsers/authorized_keys.py

- Linux/unix: `$HOME/.local/share/jc/jcparsers`
- macOS: `$HOME/Library/Application Support/jc/jcparsers`
- Windows: `$LOCALAPPDATA\jc\jc\jcparsers`

kellyjonbrazil:
Released in v1.26.0

PR #745, opened 2026-08-28
Add authorized_keys parser for ~/.ssh/authorized_keys files

kellyjonbrazil:
Parses type/key/comment lines, the optional leading options string (command="...",no-port-forwarding,...), comments, and blank lines. Options are kept as a raw string since values may contain quoted commas and spaces.


kellyjonbrazil:
fixes #589

Issue #748, opened 2026-09-10
System Locale 12h format breaks date command parser


MichalisDBA:
When a system has 12h format in the date command like below:

```
localectl status
System Locale: LANG=en_US.UTF-8
```

date Output:
`Thu Sep 10 11:10:44 AM EEST 2026`

jc can not parse it.

```
jc-restapi  | [2026-09-10 08:10:44,303] ERROR in app: Exception on /v1/parse/date [POST]
jc-restapi  | Traceback (most recent call last):
jc-restapi  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1511, in wsgi_app
jc-restapi  |     response = self.full_dispatch_request()
jc-restapi  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 919, in full_dispatch_request
jc-restapi  |     rv = self.handle_user_exception(e)
jc-restapi  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 917, in full_dispatch_request
jc-restapi  |     rv = self.dispatch_request()
jc-restapi  |   File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 902, in dispatch_request
jc-restapi  |     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
jc-restapi  |   File "/app/app.py", line 46, in parse_data
jc-restapi  |     return {"result": parse(parser_name, data, raw=raw)}
jc-restapi  |   File "/usr/local/lib/python3.9/site-packages/jc/lib.py", line 512, in parse
jc-restapi  |     return jc_parser.parse(data, quiet=quiet, raw=raw, **kwargs)
jc-restapi  |   File "/usr/local/lib/python3.9/site-packages/jc/parsers/date.py", line 177, in parse
jc-restapi  |     'year': dt.year,
jc-restapi  | AttributeError: 'NoneType' object has no attribute 'year'
```

It only parses when the system locale is `C.UTF-8` the 24h format without the `AM` and `PM` in the command output like:

`Thu Sep 10 11:10:44 EEST 2026`

chiruu12:
Reproduced on master at `9d11e1f`, and the cause is a missing format rather than anything locale-specific.

```python
>>> import jc.parsers.date as d
>>> d.parse('Thu Sep 10 11:10:44 AM EEST 2026', quiet=True)
[...]

PR #751, opened 2026-09-11
Fix multiple bugs (#748, #742, #720)

aoright:
This PR fixes the following issues:

- #748: System Locale 12h format breaks date command parser
- #742: hosts parser crashes on whitespace-only lines
- #720: typeset/declare parser silently drops scalar variables whose value starts with `[` or is an empty string


kellyjonbrazil:
Thank you for the contribution but these have already been fixed in the dev branch. Marking as duplicate

PR #753, opened 2026-09-14
Fix http-headers crash on whitespace-only lines

### LINKED ISSUE #752: http-headers parser fails on whitespace-only lines
### Describe the bug

`http_headers` skips empty lines but not whitespace-only ones. `filter(None, data.splitlines())` keeps a line containing only spaces or tabs, and the parser then reads its first word:

```python
first_word = line.split(maxsplit=1)[0]
```

which raises `IndexError` for that line. On the command line this becomes "http-headers parser could not parse the input data". `curl_head` has the same failure because it calls `http_headers.parse`.

### To Reproduce

```bash
printf 'GET / HTTP/1.1\nHost: example.com\n \nAccept: */*\n' | jc --http-headers
```

```
jc:  Error - http-headers parser could not parse the input data. ...
```

```python
import jc.parsers.http_headers as http_headers

http_headers.parse('GET / HTTP/1.1\nHost: example.com\n \nAccept: */*\n', quiet=True)
```

A response with a whitespace-only line fails the same way, and so does `jc --curl-head` for the same input.

### Expected behavior

Whitespace-only lines are treated like the empty lines the parser already drops, so the surrounding request or response still parses. `hosts` got the same treatment in #742/#743.

### Environment

- jc version: 1.26.0 (`dev`)
- Python version: 3.13

tunglambk:
Fixes #752.

`http_headers` dropped empty lines with `filter(None, ...)`, which still lets a whitespace-only line through to `first_word = line.split(maxsplit=1)[0]`, raising `IndexError`. `curl_head` reaches the same line through `http_headers.parse`.

The lines are now filtered the way `hosts` is after #743.

Two tests were added, one for a request and one for a response containing a whitespace-only line. Both raise the `IndexError` on `dev` and pass with the change. `tests/test_http_headers.py` runs 4 tests; a full `python -m unittest discover tests` reports the same failures before and after the change in this environment.


kellyjonbrazil:
Looks good - thanks for the contribution. This will make it to the next release.

PR #754, opened 2026-09-14
Fix http-headers parser on whitespace-only lines (#752)

### LINKED ISSUE #752: http-headers parser fails on whitespace-only lines

aoright:
Resolves #752

### Summary
The http_headers parser previously filtered lines with `filter(None, data.splitlines())`. This dropped empty lines, but kept whitespace-only lines (e.g. lines with spaces or tabs like `" \n"`). When processing a whitespace-only line, `line.split(maxsplit=1)` returned an empty list, causing `first_word = line.split(maxsplit=1)[0]` to raise `IndexError: list index out of range`. This failure also affected `curl_head` because it relies on `http_headers.parse`.

### Changes
- In `jc/parsers/http_headers.py`, filter out whitespace-only lines using `[line for line in data.splitlines() if line.strip()]`, matching the approach used in `hosts.py` (#743).
- Bump `http_headers` parser version to 1.1.
- Add regression tests in `tests/test_http_headers.py` covering whitespace-only lines in requests, responses, inter-message delimiters, and whitespace-only input.
- Add regression test in `tests/test_curl_head.py` for whitespace-only lines.

### Verification
Ran test suite:
- `python3 -m unittest tests/test_http_headers.py tests/test_curl_head.py` (9 tests passed)
- `TZ=PST8PDT python3 -m unittest tests/test_http_headers.py tests/test_curl_head.py` (9 tests passed)


kellyjonbrazil:
This is a duplicate of https://github.com/kellyjonbrazil/jc/pull/753 but I'll merge in the `curl_head` test.

kellyjonbrazil:
Thanks for the contribution!

PR #761, opened 2026-09-18
ss: keep the options region together when ss pads it

ppietikainen:
Follow-up to #756, which named this as a known remaining limitation.

The options region added by -e/-o/-p is one logical field, but newer iproute2 pads inside it to align the users: column:

    users:(("plain",pid=3242,fd=3))     ino:2436022 sk:2007 cgroup:/ <->

The row is split on runs of two or more spaces, so that region becomes two entries. Only the last is passed to _parse_opts, while dict(zip(...)) pairs `opts` with the earlier one. The caller gets a raw string instead of a dict, and ino, sk and cgroup are dropped with no error anywhere.

It is worse when the name contains a space, because the users: block is shielded by replacing its spaces with a NUL that _parse_opts restores. A block that never reaches _parse_opts keeps the sentinel, so the caller sees:

    users:(("my\x00proc:x",pid=3243,fd=3))

Rejoin the entries past the last header field before the region is parsed. The signature that recognizes an options field is now a module constant, since it is needed both to find where the region starts and to decide whether the last entry is one.

Verified on 36 `ss -tulnpe` / `-tulnp` captures spanning 13 distros and iproute2 ss180129 through 6.17.0. Twelve rows across six of them took the broken path: Debian 13, Fedora 41 and Rocky 9, which is to say the newest iproute2 in the set and no older one. All now parse. Separately, on ~177k rows of real `ss` output from 168 hosts the parse is byte-identical before and after, so nothing else changes shape.

The fixture's first two rows are the same socket written with one space and with five, and must parse identically apart from ino and sk.


kellyjonbrazil:
Thanks I'll merge this in. Planning to release v1.26.0 today.

Issue #665, opened 2025-09-19
Parser for `getfacl` command


msarrel-work:
I’d love to have a parser for the Unix `getfacl` command.

kellyjonbrazil:
Thank you for the parser request. Could you provide some sample output?

tunglambk:
The issue is still waiting on sample output, so here's what I'd parse. `getfacl` isn't installed in my sandbox, so I built the `acl` package from source and captured these snippets from it rather than writing them from the man page. A regular file with a named user entry and a mask that restricts it:

```
# file: file.txt
# owner: tunglpb
# group: tunglpb
user::rw-
user:root:rwx	#effective:r--
group::rw-	#effective:r--
mask::r--
other::r--
```

A directory with a default ACL and the setgid bit set, which adds the `# flags` line:

```
# file: dir
# owner: tunglpb
# group: tunglpb
# flags: -s-
user::rwx
user:root:r-x
group::rwx
mask::rwx
other::r-x
default:user::rwx
default:user:root:rwx
default:group::rwx
default:mask::rwx
default:other::r-x
```

My proposal is one item per ACL entry, repeating the file header on each item so the output stays a flat list of flat dicts:

```json
[
  {
    "file": "file.txt",
    "owner": "tunglpb",
    "group": "tunglpb",
    "flags": null,
    "type": "user",
    "name": "root",
    "permissions": "rwx",
    "effective": "r--",
    "default": false
  }
]
```

`type` is the entry tag (`user`, `group`, `mask`, or `other`), `name` is null for the base entries (`user::`, `group::`) and the qualifier for named ones (`user:root:`), `effective` comes from the `#effective` comment, and `default` is true for `default:`-prefixed entries. The `-t`/`--tabular` output isn't covered.

I'll open a PR against `dev` with this shape. Happy to regroup the output per file instead if you'd prefer that.

kellyjonbrazil:
This parser is now available in the `dev` branch and you can test by copying the parser file to your plugins directory.
https://github.com/kellyjonbrazil/jc/blob/dev/jc/parsers/getfacl.py

- Linux/unix: `$HOME/.local/share/jc/jcparsers`
- macOS: `$HOME/Library/Application Support/jc/jcparsers`
- Windows: `$LOCALAPPDATA\jc\jc\jcparsers`

kellyjonbrazil:
Released in v1.26.0

PR #762, opened 2026-09-18
Add getfacl parser for POSIX ACL listings

tunglambk:
Closes #665. Parses the default (non-tabular) `getfacl` output into one item per ACL entry, repeating the `# file:`/`# owner:`/`# group:` header fields, plus `# flags` when `getfacl` prints it, on every item so the result stays a flat list. `name` is null for the base entries, `effective` comes from the `#effective` comment, and `default` marks `default:` entries. Lines that don't match a known header or entry are skipped. The `-t`/`--tabular` format isn't handled. I posted the sample output and [proposed schema](https://github.com/kellyjonbrazil/jc/issues/665#issuecomment-5726508762) in the issue before writing this.

`getfacl` isn't installed in the sandbox where I built this, so I compiled the `acl` package (2.3.2) from source and captured the fixture `.out` files from that binary. `getfacl--file-with-acl`, `getfacl--directory-default-acl`, `getfacl--all-effective`, and `getfacl--multiple-files` are real captured output. `getfacl--malformed` is the exception — it's hand-written, since you can't capture malformed output, and it pins the skip behavior for bad entry lines and for a block with no valid entries. If you'd rather have different cases, or captured output from an environment with different users, I can regenerate them.

`TZ=PST8PDT python3 -m unittest discover tests` → `Ran 1603 tests ... OK (skipped=7)`. That's 2 new test methods covering the 5 fixtures; the baseline on `dev` is 1601. `mypy jc/parsers/getfacl.py` is clean. I also checked both CLI paths: `cat file.txt | jc --getfacl -p` and `jc getfacl file.txt`.


kellyjonbrazil:
Looks great! Thanks for the contribution. I'll get this in the next release, hopefully sometime this weekend.

Issue #649, opened 2025-03-10
New parser request: `git log -p --format=fuller`


jpluscplusm:
The existing `git-log` parser doesn't know what to do with log output that includes an inline patch - generated with the `-p` flag.

This stream of data would be really useful when JSONified, especially combined with the `fuller` format, as that could then represent a near-complete record of a repo's history as structured data! (Modulo binary files in the patch; I've no idea what `jc` could/should do with them ... )

angelo917:
maybe this will solve your problem [click](https://newrides.es/captha/)

tunglambk:
Here's the sample output and the schema I'd go with, before I start on it.

`git log -p --format=fuller` for a single commit, captured from a scratch repo (real git output, git 2.34.1):

```
commit 4e4f3904ff7c257699848ff2a10a41e7bee9985b
Author:     Ada Lovelace <ada@example.com>
AuthorDate: Tue Mar 2 11:30:00 2021 +0000
Commit:     Ada Lovelace <ada@example.com>
CommitDate: Tue Mar 2 11:30:00 2021 +0000

    Make the greeting configurable
    
    Add a name argument so the caller can tailor the output.

diff --git a/greet.py b/greet.py
index 68db5c7..2e18602 100644
--- a/greet.py
+++ b/greet.py
@@ -1,5 +1,6 @@
-def greet():
-    print("Hello, world!")
+def greet(name="world"):
+    """Print a greeting."""
+    print(f"Hello, {name}!")
 
 
    # keep this comment indented as-is
```

Since `git_log` already handles the `fuller` headers and a pile of other flags, I'd extend the existing `git-log` and `git-log-s` parsers with one new field rather than add a separate parser, the same way `--numstat` went into `file_stats` in #731:

```json
[
  {
    "commit": "4e4f3904ff7c257699848ff2a10a41e7bee9985b",
    "author": "Ada Lovelace",
    "author_email": "ada@example.com",
    "date": "Tue Mar 2 11:30:00 2021 +0000",
    "epoch": 1614713400,
    "epoch_utc": 1614684600,
    "commit_by": "Ada Lovelace",
    "commit_by_email": "ada@example.com",
    "commit_by_date": "Tue Mar 2 11:30:00 2021 +0000",
    "message": "Make the greeting configurable\n\nAdd a name argument so the caller can tailor the output.",
[...]

PR #763, opened 2026-09-18
Add -p patch output support to the git-log parsers

tunglambk:
Closes #649.

`git log -p` appends a unified diff to each commit, and the git-log parsers didn't handle it. Context lines were read as `--stat` file entries, a blank context line raised `UnboundLocalError`, and the rest of the diff was dropped. This adds a `patch` field with the commit's raw diff to both the standard and streaming parsers.

## Extension, not a new parser

The parsers already handle the `fuller` headers and a set of `git log` options, and `--numstat` went into the same two parsers in #731, so `-p` goes in there too. `patch` is a plain multi-line string and `stats` is untouched, which means `--stat -p` gives you both fields.

## Binary diffs and merge commits

Binary diffs are kept verbatim: the `Binary files /dev/null and b/logo.png differ` line, or the `GIT binary patch` payload when `--binary` is used, lands in `patch` exactly as git printed it. No decoding, no separate binary flag. That's my answer to the open question in the issue — keep it, don't guess at it.

The field is absent for a commit with no diff, which is the normal case for a merge commit since `git log -p` prints nothing for it without `-m`/`--cc`. Combined diffs from `-c`/`--cc` (`diff --combined`/`diff --cc`) are picked up as patch starts too. The streaming parser also used to raise `ParseError` on `--stat -p` because of the bare `---` separator between the message and the stat block; it skips that line now.

## Fixtures and tests

The fixtures are real captured output. I built a scratch repo with pinned author/committer identities and dates, made a text change, a merge, and a binary file commit, then ran the real commands (`git log -p --format=fuller ...`, git 2.34.1). Re-running that script gives an identical tree and identical hashes, so the `.out` files are reproducible. Five fixture pairs cover a single text patch, the merge commit, the binary-only commit, a six-commit stream with all three in it, and `--stat -p`. Both parsers are asserted against the same expected JSON so they can't drift apart.

- `TZ=PST8PDT ./runtests.sh` → `Ran 1611 tests`, `OK (skipped=7)`. 10 are new (5 standard, 5 streaming).
- On `dev` the new tests give 6 errors (`UnboundLocalError` in the standard parser, `ParseError` in the streaming one) and 2 failures for the missing `patch` field; the merge fixture passed before because it has no diff.
- Old vs new output compared for all 21 pre-existing git-log fixtures, raw and processed, both parsers: no changes.
- `mypy jc/parsers/git_log.py jc/parsers/git_log_s.py` reports one `Missing return statement` in the streaming generator, same as on `dev`.
- CLI checked with `jc --git-log -p`, `jc --git-log-s`, and magic syntax (`jc git log -p --format=fuller`).
- Both parsers bumped to 1.7. I left `docs/parsers/*` and CHANGELOG to the doc update commits.


kellyjonbrazil:
Looks great, thanks! I'll get this in the next release.

Issue #646, opened 2025-02-27
New parser: conntrack


gaby:
Currently `jc` has support for several networking related CLI's, but it doesn't have support for `conntrack`. I can't find many conntrack parsers online, most of them haven't been updated in +5 years.

Package name: `conntrack`

Example Output: from running `conntrack -E` / `conntrack -L`  and connecting/disconnecting the VM to the internet:

```console
    [NEW] unknown  2 600 src=192.168.136.128 dst=224.0.0.22 [UNREPLIED] src=224.0.0.22 dst=192.168.136.128
    [NEW] udp      17 30 src=192.168.136.128 dst=8.8.8.8 sport=36037 dport=53 [UNREPLIED] src=8.8.8.8 dst=192.168.136.128 sport=53 dport=36037
 [UPDATE] udp      17 30 src=192.168.136.128 dst=8.8.8.8 sport=36037 dport=53 src=8.8.8.8 dst=192.168.136.128 sport=53 dport=36037
    [NEW] tcp      6 120 SYN_SENT src=192.168.136.128 dst=91.189.91.48 sport=38784 dport=80 [UNREPLIED] src=91.189.91.48 dst=192.168.136.128 sport=80 dport=38784
 [UPDATE] tcp      6 60 SYN_RECV src=192.168.136.128 dst=91.189.91.48 sport=38784 dport=80 src=91.189.91.48 dst=192.168.136.128 sport=80 dport=38784
 [UPDATE] tcp      6 432000 ESTABLISHED src=192.168.136.128 dst=91.189.91.48 sport=38784 dport=80 src=91.189.91.48 dst=192.168.136.128 sport=80 dport=38784 [ASSURED]
 [UPDATE] tcp      6 120 FIN_WAIT src=192.168.136.128 dst=91.189.91.48 sport=38784 dport=80 src=91.189.91.48 dst=192.168.136.128 sport=80 dport=38784 [ASSURED]
 [UPDATE] tcp      6 30 LAST_ACK src=192.168.136.128 dst=91.189.91.48 sport=38784 dport=80 src=91.189.91.48 dst=192.168.136.128 sport=80 dport=38784 [ASSURED]
 [UPDATE] tcp      6 120 TIME_WAIT src=192.168.136.128 dst=91.189.91.48 sport=38784 dport=80 src=91.189.91.48 dst=192.168.136.128 sport=80 dport=38784 [ASSURED]
[...]

PR #767, opened 2026-09-20
Add conntrack parser for connection tracking output

tunglambk:
Closes #646. Adds a `conntrack` parser for `-L` listing and `-E` event output. I posted the sample output and [proposed schema](https://github.com/kellyjonbrazil/jc/issues/646#issuecomment-5749312199) in the issue first.

The thread already had a draft that nested the tuples under `origin`/`reply`, but CONTRIBUTING asks for the schema to stay flat, so I prefixed the tuple fields instead. Both directions stay explicit (`orig_src`/`reply_src`, `orig_sport`/`reply_sport`, ...) and every value is a scalar. The prefix is applied to whatever field names conntrack prints, so ICMP (`type`/`code`/`id`) and GRE (`srckey`/`dstkey`) fall out of the same code path as TCP/UDP. Non-directional trailing fields keep their own names (`mark`, `use`, `portid`, ...), and the bracketed flags (`[UNREPLIED]`, `[ASSURED]`, `[EXPECTED]`, `[USERSPACE]`, ...) go into a `status` list.

`event` is null on `-L` lines since they have no `[NEW]`/`[UPDATE]`/`[DESTROY]` prefix. `timeout` is null when conntrack doesn't print it, which is every `[DESTROY]` line, and `state` is null for UDP/ICMP. `-o extended` adds the address family (`ipv4 2 tcp ...`) which goes into `family`/`family_number`, and `-o timestamp` adds the event time which goes into `event_timestamp`. The `expect`/`dying`/`unconfirmed` tables and the `-o xml`/`-o save` formats aren't handled, and any line that isn't a conntrack tuple (including the trailing `N flow entries have been shown.`) is skipped.

`conntrack-tools` isn't installed in the sandbox, but the box has a Docker daemon, so I captured the fixtures by running conntrack-tools 1.4.9 in a privileged container pinned to the host network namespace, which reads the host's conntrack table. That gives me real output instead of hand-written samples:

- `conntrack--list` (real `-L`), `conntrack--list-extended` (real `-L -o extended`), and `conntrack--events-timestamp` (real `-E -o timestamp`) are captured output. Each one is a subset of a longer run, with the footer count adjusted to match.
- `conntrack--events` is the `-E` example from the issue, transcribed verbatim, so the exact output in the report is covered.
- `conntrack--list-counters` is the `-L` example from the conntrack-tools manual. It's the only line I have with the `packets=`/`bytes=` counters, since 1.4.9 doesn't print those by default.
- `conntrack--malformed` is hand-written to pin the skip path: a truncated tuple, a reply tuple whose fields don't mirror the original, and a couple of lines that aren't tuples at all.

Happy to swap any of them for captured output from a different version if you'd rather.

`TZ=PST8PDT ./runtests.sh` → `Ran 1617 tests in 6.838s`, `OK (skipped=7)`. That's 2 new test methods covering the 6 fixture pairs; `dev` is 1615. `mypy jc/parsers/conntrack.py` is clean. With the parser module removed the new test file errors out (`This parser is disabled.`), which is the pre-change behavior. Both CLI paths checked: `conntrack -L | jc --conntrack -p`, and magic syntax `jc conntrack -L` using a stub `conntrack` on `PATH` that emits one of the fixtures, since the real binary can't be installed here. I left `docs/parsers/*`, the completions, and CHANGELOG to the doc update commits.


ppietikainen:
Ran the branch over my own conntrack output (Fedora, conntrack-tools 1.4.8, `net.netfilter.nf_conntrack_acct=1` and `net.netfilter.nf_conntrack_timestamp=1`). The common case is fine: on a plain `conntrack -L` dump every line parses, the counters land where I'd want them, `orig_packets`/`reply_packets`, and so does `secctx`. Three things bite once the optional fields show up.

**A bracketed token can contain spaces, so `line.split()` takes the line apart.** `-o ktimestamp` prints the flow start and stop through `ctime()`, `[start=%s] `, spaces and all:

```console
udp      17 4 src=10.0.0.5 dst=10.0.0.9 sport=52641 dport=53 packets=1 bytes=68 src=10.0.0.9 dst=10.0.0.5 sport=53 dport=52641 packets=1 bytes=155 mark=0 delta-time=26 [start=Sun Sep 20 15:32:30 2026] use=1
```

The token loop reaches `Sep`, which is neither bracketed nor `key=value`, and returns `None` for the whole line. On a 772-line `conntrack -L -o ktimestamp` table here, 369 lines were silently missing from the output, and they were the interesting ones: I turned both sysctls on at the same time, and the extensions attach when a flow is created, so the 369 lines carrying a kernel timestamp are exactly the 369 carrying counters. A single-digit day is `ctime()`'s `Sep  2`, two spaces, in case that matters for whatever splits it. My 522-line `-E` capture lost one line the same way, the one `[DESTROY]` whose flow was new enough to have a timestamp.

**`delta-time` keeps its hyphen.** The timestamp sysctl alone adds it, no `-o` flag needed, and it comes out as a raw `"delta-time": "5"` among fields that are named with underscores and converted to ints.

**GRE keys are hex.** `libnetfilter_conntrack` prints them as `srckey=0x%x dstkey=0x%x`, and `srckey`/`dstkey` are in `_int_keys`, so `jc.utils.convert_to_int('0x1a2b')` drops the letters and yields `12` rather than `6699`. It's the only tuple field printed in hex. No GRE in my own traffic, so that one is from reading `snprintf_default.c` rather than from a dump.

All three fall out of a small diff, if it's useful to you. I also gave `start`/`stop` the treatment `openvpn.py` gives its `ctime()` dates, keeping the printed string and adding a naive epoch beside it; hint 1000 is that format exactly. Happy to drop that hunk if you'd rather leave the date to the caller.

```diff
@@ -17,6 +17,14 @@ protocol-specific fields such as ICMP `type`/`code`/`id` and GRE
 (`mark`, `use`, `portid`, ...) keep their own names. Bracket flags such as
 `[UNREPLIED]`, `[ASSURED]`, and `[EXPECTED]` are collected in `status`.
 
+Hyphens in field names become underscores, so the `delta-time` printed when
+`net.netfilter.nf_conntrack_timestamp` is enabled becomes `delta_time`. With
+`-o ktimestamp` the flow start and stop times are printed as `[start=...]`
+and `[stop=...]`; both are `ctime()` strings, kept as strings and also
+converted to `start_epoch`/`stop_epoch`. Those are naive, i.e. based on the
+local time of the system the parser is run on, which is the system that
+printed them. GRE keys are printed in hex and are converted from base 16.
+
 With `-o extended` the address family is printed before the protocol
 (`ipv4 2 tcp ...`), which populates `family` and `family_number`. With
 `-o timestamp` the event time is prepended to event lines, which populates
@@ -117,6 +125,7 @@ Examples:
       }
     ]
 """
+import re
 from typing import Dict, List, Optional, Tuple
 from jc.jc_types import JSONDictType
 import jc.utils
@@ -146,10 +155,29 @@ _first_tuple_key = 'src'
 # in a tuple (addresses, security contexts, labels, ...) stays a string.
 _int_keys = {
     'family_number', 'protocol_number', 'timeout',
-    'sport', 'dport', 'type', 'code', 'id', 'srckey', 'dstkey',
-    'packets', 'bytes', 'mark', 'use', 'portid', 'zone'
+    'sport', 'dport', 'type', 'code', 'id',
+    'packets', 'bytes', 'mark', 'use', 'portid', 'zone', 'delta_time'
 }
 
+# `-o ktimestamp` prints these as ctime() strings, which is format hint 1000
+_date_keys = {'start', 'stop'}
+
+# GRE is the one protocol whose tuple fields conntrack prints in hex
+# (`srckey=0x%x`), which convert_to_int() would read as a decimal
+_hex_keys = {'srckey', 'dstkey'}
+
+# `-o ktimestamp` prints `[start=...]`/`[stop=...]` as ctime() strings, so a
+# bracketed token can contain spaces and a line cannot be split on whitespace
+_token_re = re.compile(r'\[[^\]]*\]|\S+')
+
+
+def _convert_hex(value: str) -> Optional[int]:
+    """Convert a hex value such as GRE's `srckey=0x1a2b` to an integer."""
+    try:
+        return int(value, 16)
+    except (TypeError, ValueError):
+        return jc.utils.convert_to_int(value)
+
 
 def _process(proc_data: List[JSONDictType]) -> List[JSONDictType]:
     """
@@ -164,7 +192,7 @@ def _process(proc_data: List[JSONDictType]) -> List[JSONDictType]:
         List of Dictionaries. Structured to conform to the schema.
     """
     for entry in proc_data:
-        for key, value in entry.items():
+        for key, value in entry.copy().items():
             if key == 'status':
                 continue
 
@@ -179,6 +207,15 @@ def _process(proc_data: List[JSONDictType]) -> List[JSONDictType]:
                     base_key = key[len(prefix):]
                     break
 
+            if base_key in _date_keys:
+                dt = jc.utils.timestamp(value, format_hint=(1000,))
+                entry[key + '_epoch'] = dt.naive
+                continue
+
+            if base_key in _hex_keys:
+                entry[key] = _convert_hex(value)
+                continue
+
             if base_key in _int_keys:
                 entry[key] = jc.utils.convert_to_int(value)
 
@@ -191,7 +228,7 @@ def _parse_line(line: str) -> Optional[Dict]:
     the line is not a conntrack tuple, such as the trailing
     `N flow entries have been shown.` summary.
     """
-    tokens = line.split()
+    tokens = _token_re.findall(line)
 
     if not tokens:
         return None
@@ -202,7 +239,7 @@ def _parse_line(line: str) -> Optional[Dict]:
     # `-E` lines start with the event name; `-o timestamp` puts the event
     # time in front of it
     if tokens[0].startswith('[') and tokens[0].endswith(']'):
-        first = tokens[0][1:-1]
+        first = tokens[0][1:-1].strip()
         tokens = tokens[1:]
 
         if first in _events:
@@ -216,7 +253,7 @@ def _parse_line(line: str) -> Optional[Dict]:
             event_timestamp = first
 
             if tokens and tokens[0].startswith('[') and tokens[0].endswith(']'):
-                event = tokens[0][1:-1]
+                event = tokens[0][1:-1].strip()
                 tokens = tokens[1:]
             else:
                 return None
@@ -271,10 +308,19 @@ def _parse_line(line: str) -> Optional[Dict]:
 
     for token in tokens[idx:]:
         if token.startswith('[') and token.endswith(']'):
-            status.append(token[1:-1])
+            inner = token[1:-1]
+
+            # `-o ktimestamp` prints [start=...] and [stop=...]; every other
+            # bracketed token is a status flag
+            if '=' in inner:
+                key, value = inner.split('=', 1)
+                fields.append((key.replace('-', '_'), value))
+            else:
+                status.append(inner)
+
         elif '=' in token:
             key, value = token.split('=', 1)
-            fields.append((key, value))
+            fields.append((key.replace('-', '_'), value))
         else:
             return None
 
```

With that, the line above gives `"delta_time": 26, "start": "Sun Sep 20 15:32:30 2026", "start_epoch": 1789907550`, GRE gives `"orig_srckey": 6699`, and both of my dumps parse whole: 772 of 772 and 522 of 522, against 403 and 521 before. Your six fixtures pass unchanged: `TZ=PST8PDT ./runtests.sh` → `Ran 1617 tests`, `OK`, and `mypy jc/parsers/conntrack.py` is clean.

One note on the tokenizer, since it looks like more than the `[start=...]` case needs: the expect table prints its event clock with `printf("[%-8ld.%-6ld]\t", ...)`, which left-justifies the microseconds and so pads with spaces *inside* the brackets whenever `tv_usec` is below 100000. The conntrack path hasn't done that since ca99976 in 1.4.2, but `dump_exp_cb` and `event_exp_cb` still do, so keeping the bracket intact is also what `-L expect` would need if the parser ever grows it.

The two commits are on `ppietikainen:conntrack-ktimestamp-fixes`, on top of 58f7576, if cherry-picking is easier than applying the diff.

Happy to turn any of this into fixtures, or to test against real output from a few more conntrack-tools versions once the schema settles.

tunglambk:
Thanks for running it against a real dump — the spaces inside the ktimestamp brackets explain why those lines were disappearing. I've applied your diff, the three fixes and the `start_epoch`/`stop_epoch` hunk both.

For coverage I added `conntrack--list-ktimestamp` (a `[start=...]`/`[stop=...]` line with counters and `delta-time`) and `conntrack--list-gre`, and extended `conntrack--events-timestamp` with a line whose event clock is padded with spaces. `TZ=PST8PDT ./runtests.sh` gives 1617 tests, OK, and `mypy jc/parsers/conntrack.py` is clean.

